PR #316, opened 2014-12-15
made the CLOUD_MODE demo user account a setting

allo-:
Option which account should be the demo account.


xiez:
Merged. Sorry for the late response.

PR #334, opened 2015-01-05
Update django.po

dervomsee:
Wrong german Translation!
"Öffentlich teilen" means "share public". The translation "Datei herunterlaten" would be better.


nafets:
Yes, without a typo would be great:

```
msgstr "Datei herunterladen"
```

Tischi86:
You can also download directorys with this translations.
In my mind this translation is better https://github.com/haiwen/seahub/pull/341

PR #341, opened 2015-01-10
Fixed german "download" translation

Tischi86:
"Öffentlich teilen" makes no sense here.
You can look to http://www.dict.cc/?s=download how to translate "download"


Tischi86:
fixed yet

PR #353, opened 2015-01-27
[repo-setting] Set folder perm

imwhatiam:


seahub/views/ajax.py
```diff
@@ -1859,3 +1930,355 @@ def events(request):
...
+
+    content_type = 'application/json; charset=utf-8'
+    username = request.user.username
+
+    repo = seafile_api.get_repo(repo_id)
+    if not repo:
+        raise Http404
+
+    # no settings for virtual repo
+    if ENABLE_SUB_LIBRARY and repo.is_virtual:
```
freeplant: ENABLE_SUB_LIBRARY is always true. So checking this variable is not necessary.

Issue #392, opened 2015-03-04
Download a library


jeremejevs:
In my case, I want to be able to download a shared directory. Right now it's treated as a library, so there's no download button in the interface, but if I just append the ID to `/seafile/repo/download_dir/`, then I'm able to download even raw libraries.

My current workaround is to add the following to repo buttons in `my_owned_repos.html`:

``` html
<a href="{% url 'repo_download_dir' repo.props.id %}" class="repo-download-btn op-icon vh" title="{% trans "Download" %}">
  <img src="{{MEDIA_URL}}img/download.png" alt="" />
</a>
```

And then a slightly different version to repo buttons in `my_shared_repos.html` and `my_group_repos.html`:

``` html
<a href="{% url 'repo_download_dir' repo.repo_id %}" class="repo-download-btn op-icon vh" title="{% trans "Download" %}">
  <img src="{{MEDIA_URL}}img/download.png" alt="" />
</a>
```

It would also be nice to have a big download button inside directories, like this one, added to top bar buttons in `repo_dir_data.html`:

``` html
<button onclick="location.href='{% url 'repo_download_dir' repo.id %}?p={{ path|urlencode }}'" class="op-btn">
  <span class="icon-download"></span>
  {% trans "Download"%}
</button>
```

freeplant:
We don't have a plan to implement such feature.

PR #393, opened 2015-03-04
Add more download buttons

jeremejevs:
As per haiwen/seahub#392. My solution is somewhat hack-ish, so I expect it to be cleaned up before being merged. For example, a button doing `location.href` on click is certainly unacceptable, but that was the only way to achieve what I wanted without modifying CSS. Also, there should probably be an `.icon-download-alt` rule, using an [appropriate download icon](http://fortawesome.github.io/Font-Awesome/icon/download/), and it would probably make more sense to generate download links in controllers, similar to how it's done in `repo_dirents.html`.

### DIFF
--- a/seahub/templates/snippets/my_group_repos.html
+++ b/seahub/templates/snippets/my_group_repos.html
@@ -21,6 +21,7 @@
             <td></td>
           {% endif %}
             <td>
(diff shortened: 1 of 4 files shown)


freeplant:
It will be added in v4.2.

PR #395, opened 2015-03-04
Shibboleth variable should be configurable. 

neffs:
Environment variable for shibboleth user name should  be configurable. The default is REMOTE_USER which is provided by e.g. apache modules. Using a different variable the value can be changed, e.g. a Domain can be added or replaced or a different field can be used as a primary key. 

This allows to use other authentication modules for Apache, like Kerberos or CAS which don't necessarily provide a username in the right format (xxx@yyy.com). 

Tested using LemonLDAP::NG Single-Sign-On.


xiez:
Thanks!

PR #397, opened 2015-03-06
Backbone shared lib

freeplant:


media/scripts/common.js
```diff
@@ -75,6 +75,8 @@ define([
               case 'cancel_cp': return siteRoot + 'ajax/cancel_cp/';
               case 'get_shared_link': return '';
               case 'get_shared_upload_link': return '';
+
+              case 'ajax_repo_leave_share': return siteRoot + 'share/ajax/repo_leave_share/';
```
freeplant: Need change to ajax_repo_remove_share too

seahub/share/views.py
```diff
@@ -270,6 +270,87 @@ def share_repo(request):
 
     return HttpResponseRedirect(next)
 
+@login_required_ajax
+def ajax_repo_leave_share(request):
```
freeplant: need change to 'ajax_repo_remove_share'

PR #408, opened 2015-03-10
[api] support logging out an accout from seafile clients

lins05:
Added new apis:

- Logout a client: `/logout-device/`. If the device is a desktop client, also remove all the sync tokens used on that client.
- Generate sync tokens for a list repos: `repo-tokens`. This is used when the user logs in the account again in the client.

Related:

- haiwen/seafile-client#476
- haiwen/seafile#1112

### DIFF
--- a/seahub/api2/views.py
+++ b/seahub/api2/views.py
@@ -5,6 +5,7 @@ import stat
 import json
 import datetime
 import urllib2
+import re
 from urllib2 import unquote, quote
 from PIL import Image
 from StringIO import StringIO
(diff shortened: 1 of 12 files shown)


seahub/api2/views.py
```diff
@@ -3573,6 +3575,35 @@ def get(self, request, repo_id):
...
+    def get(self, request, format=None):
+        repos = request.GET.get('repos', None)
+        if not repos:
+            return api_error(status.HTTP_400_BAD_REQUEST, "You must specify libaries ids")
+
+        repos = [repo for repo in repos.split(',') if repo]
+        if any([not _REPO_ID_PATTERN.match(repo) for repo in repos]):
+            return api_error(status.HTTP_400_BAD_REQUEST, "Libraries ids are invalid")
+
+        if any([not seafile_api.check_repo_access_permission(
```
freeplant: What if a user no longer have a permission to a repo after logout? Then he will not be able to login again?

lins05: That could be a problem. The same may also happen when a library is deleted. 

I can think of two options here:

- Option 1: only return tokens for repos the user can access, and simply ignore those the user can't access. Dirty and quick.
- Option 2: tell the client the server failed to access some repos, like this:

```python
{
    'ok': {
        'repo_1': 'token_1'
        'repo_2': 'token_2'
    },
    'failed': {
        'repo_3': 'access denied',
        'repo_4': 'no such repo',
    }
}
```

PR #417, opened 2015-03-18
share

llj:


seahub/share/views.py
```diff
@@ -1268,13 +1277,294 @@ def send_shared_upload_link(request):
...
+        username = request.user.username
+
+        if not path.endswith('/'):
+            path = path + '/'
+
+        l = UploadLinkShare.objects.filter(repo_id=repo_id).filter(
+            username=username).filter(path=path)
+        if len(l) > 0:
+            token = l[0].token
+            print token
```
freeplant: This line should be removed.

seahub/share/views.py
```diff
@@ -1268,13 +1277,294 @@ def send_shared_upload_link(request):
...
+            share_to_users.append(email)
+        else:
+            shared_failed.append(email)
+
+    for group_id in groups:
+        share_to_groups.append(seaserv.get_group(group_id))
+
+    if not check_user_share_quota(username, shared_repo, users=share_to_users,
+                                  groups=share_to_groups):
+        result['error'] = _(('Failed to share "%s", no enough quota. <a href="http://seafile.com/">Upgrade account.</a>') % shared_repo.name)
```
freeplant: should escape the shared_repo.name

seahub/share/views.py
llj: 不是我写的。

PR #419, opened 2015-03-19
Enable create repo at myhome and group pages

xiez:


media/scripts/app/models/repo.js
```diff
@@ -32,6 +32,10 @@ define([
             if (attrs.encrypted) {
                 if (!attrs.passwd1)  return gettext("Please enter password");
                 if (!attrs.passwd2)  return gettext("Please enter the password again");
+                if (attrs.passwd1.length < app.pageOptions.repo_password_min_length) {
+                    return gettext("Password is too short");
+
```
freeplant: The empty line should be removed

PR #437, opened 2015-03-25
Redirect using "next" parameter when using shibboleth button

neffs:
Use the next parameter for shibboleth, just like for the normal login button.


lins05:
LGTM.

PR #439, opened 2015-03-26
implement the <time> html5 tag for all times.

cbergmann:
I implemented the change like I proposed in Ticket haiwen/seafile#1146 .
I am not fluent enough neither in python nor in django to know if we have to do something about Timezones at this point.
The Diff is a little bit confused because I renamed the old Function and inserted a function with the same name. One can see the code more clear in the split view or the file view.


xiez:
Modified (https://github.com/haiwen/seahub/commit/5746ef946522f990a5f5f62da5a21524613a5411) and merged!  Thanks @cbergmann, really appreciate it!

PR #442, opened 2015-03-29
Add Japanese translation

miurahr:
*Add Japanese locale
*Add Japanese translation, contributions and sync with transifex 

Signed-off-by: Hiroshi Miura <miurahr@linux.com>


xiez:
Thanks! @miurahr

PR #476, opened 2015-04-10
[thumbnail] add http cache control

imwhatiam:


seahub/thumbnail/views.py
```diff
@@ -134,8 +148,14 @@ def thumbnail_get(request, repo_id, obj_id, size=THUMBNAIL_DEFAULT_SIZE):
     thumbnail_file = os.path.join(THUMBNAIL_ROOT, size, obj_id)
     try:
         with open(thumbnail_file, 'rb') as f:
-            thumbnail = f.read()
+            file_content = f.read()
         f.close()
```
xiez: 这个不需要close吧

imwhatiam: 在python的文档中 是这样说的
It is good practice to use the with keyword when dealing with file objects. This has the advantage that the file is properly closed after its suite finishes, even if an exception is raised on the way. It is also much shorter than writing equivalent try-finally blocks:
```
with open('workfile', 'r') as f:
    read_data = f.read()
f.closed
```

PR #506, opened 2015-04-20
[thumbnail] use 'image/png' as content-type

imwhatiam:


lins05:
> THUMBNAIL_EXTENSION = 'png'

What if the use changes the extension type?

PR #507, opened 2015-04-20
Bugfix creating a public share set_inner_pub_repo does not exist in Seaf...

joshbmarshall:
...ileAPI

Previously the Web API call to share a library as public would give an error in the log and fail. Renaming the call from set_inner_pub_repo to add_inner_pub_repo works perfectly


xiez:
Thanks! @joshbmarshall

PR #529, opened 2015-04-27
support logging in from desktop client directly

lins05:
- New table: `base.ClientLoginToken`
- New api endpiont: `/api2/client-login/`, for generating the login token
- New view: `/client-login/`, for checking the token and logging in the session


lins05:
This PR should go to the master (4.2) branch, since it adds a new table.

PR #531, opened 2015-04-27
support logging in from desktop client directly

lins05:
- New table: `base.ClientLoginToken`
- New api endpiont: `/api2/client-login/`, for generating the login token
- New view: `/client-login/`, for checking the token and logging in the session

### DIFF
--- a/seahub/api2/views_auth.py
+++ b/seahub/api2/views_auth.py
@@ -29,9 +29,11 @@ class LogoutDeviceView(APIView):
         return {}
 
 class ClientLoginTokenView(APIView):
-    """Removes the api token of a device that has already logged in. If the device
-    is a desktop client, also remove all sync tokens of repos synced on that
-    client .
+    """Generate a token which can be used later to login directly.
+
+    This is used to quickly login to seahub from desktop clients. The token
+    can only be used once, and would only be valid in 30 seconds after
+    creation.
     """
     authentication_classes = (TokenAuthentication,)
     permission_classes = (IsAuthenticated,)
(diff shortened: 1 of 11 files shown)


seahub/api2/views_auth.py
```diff
@@ -24,3 +27,19 @@ def post(self, request, format=None):
             seafile_api.delete_repo_tokens_by_peer_id(request.user.username, auth_token.device_id)
         auth_token.delete()
         return {}
+
+class ClientLoginTokenView(APIView):
+    """Removes the api token of a device that has already logged in. If the device
+    is a desktop client, also remove all sync tokens of repos synced on that
```
freeplant: This comment is not right.

lins05: @freeplant updated.

shoeper:
I'd recommend to restrict access for automatically created login tokens. E.g. admin page should only be allowed with password, I'd also recommend to ask for the password again after few minutes (option in config, 0 mins = unlimited everything else is the time until one is asked again). Furthermore make sure to only allow account deletion after the password has been verified.

freeplant:
Yes, it is better to require handle admin pages specially (ask password for a second time).

lins05:
@freeplant PR updated. Now if the admin logs in to seahub from the desktop client, he would still be asked to confirm password when entering sysadmin area.

Ref #544

shoeper:
Nice that you've implemented it.
Imho just on case left: Could be nice to also enable it for "login for 7 days" (but the 7days login is not so important anymore when one is directly logged after opening seafile via client) .

PR #542, opened 2015-04-30
Fix onetime token bug in video/audio bug

xiez:


shoeper:
Are you sure you only want to enable byte range requests for video and audio?

Think of cases like e.g. one wants to download a 50 MiB file offsite while having a very bad connection. With byte range requests, one would be able to continue the file download. Without one has to try again until the the connection is not getting interrupted while downloading. This does not only take more time but also wastes mobile data volume.

freeplant:
It only affects whether the access token (the random URL) can be used for more than once. The random URLs for files always support byte range request. But most tools will not be able to do resumable download if redirected to a different URL. Mobile clients written by ourselves can correctly handle transfer resuming using a different access token.

shoeper:
But byte range requests require multiple token usages, don't they?

freeplant:
The workflow is as following:

1. When a user click a download button for a file, a random url will be created to access the file like https://cloud.seafile.com/seafhttp/files/142e3a6e-305e-4624-87e6-3c326eec41f9/001.png
2. The address can be set to visited only once (the default behavior) https://cloud.seafile.com/seafhttp/files/142e3a6e-305e-4624-87e6-3c326eec41f9/001.png
3. The address itself is support byte range request. You can visit the file start at 1MB offset.

shoeper:
So if I understand right I can not start file download with that link an make another request to continue the download at byte y without getting a new link.

freeplant:
You can continue the download at that byte by getting a new link for that file.

shoeper:
Yeah, that's the problem I am talking about. This practically breaks the ability to continue downloads.

Issue #594, opened 2015-05-13
Downscaled Images for the new gallery view


dhoffend:
While I really like the new gallery view to walk through images, I dislike it the same way. It's using the original image which can several MB in size. Walking through 10-20 images can already take 100MB since the usual pictures taken from modern cameras are between 4-13 MB per picture.

A good way could be to provide reduced/compressed images with 2048px of the larges side. Maybe we could use the PREVIEW_DEFAULT_SIZE parameter for this provide within the dirent list additionally to the thumbnail_src a preview_src that points to the big preview images (which should be used the the modal image viewer which can be smaller in size (300-600kb instead of 6mb).

The only concern I would have would be security about the thumbnail urls (are they persistent urls?).

Btw. for the gallery feature of 5.1 (or whenever) it might be useful to have 3 sizes anyway: thumbnail (32x32), grid (256x256), preview (2048x2048) but that's a different story.

dhoffend:
We could use the ajax content function of the magnificPopup plugin to generate the downscaled images on demand.

btw. the preview box for shared folders should be replaced by this popup also

dhoffend:
I'm working on it. Got something working using the elementParse callback of the magnificPopup Plugin.

Works great so far, i'll send a pull request once i'm done with cleaing up the python code. I also had to tweak the thumbnail_create.

dhoffend:
Addition stuff for the future: remote PREVIEW images from the shared directory view and replace them with the lightbox

dhoffend:
Oh and the image file preview as well. IMHO/Maybe the fullsize image should only be served for a download request.

dhoffend:
One question to the original authors: is there a garbage collection implimented to get rid of outdated thumbnails or is the best way to enable atime (accesstime) stats on the filesystems and remove those that have not been used for time X?

xiez:
there is no GC for thumbnail yet, we can provide a django management command or just a shell script to clean thumbnails have not been used for time X.

dhoffend:
But this would require the atime attribute set on the filesystem right? I guess the same applies for images that have been deleted on seafile or modified? But I guess modified means new object_id which results in a new image i guess

PR #602, opened 2015-05-18
Enable thumbnail generation for image lightbox

dhoffend:
This is a pull request based on the ticket: #594

This update makes use of the thumbnail generation for larger image
galleries. While browsing images using the lightbox the user
doesn't have to download the full size image which could have several
mb in size. instead a downscaled image is created (default 1280px).

The thumbnail_create function is modified to use jpeg instead of png
for thumbs > 100 pixel to save space (1mb > 200kb).

The generation hooks into the mangificPopop elementParse() callback
but as a blocking request until the thumb is generated.

The new settings variables have been introduced:
- THUMBNAIL_LARGE = '100'
- THUMBNAIL_EXTENSION_LARGE = 'jpeg'
- ENABLE_THUMBNAIL_LARGE = True
- THUMBNAIL_LARGE_SIZE = '1280'

A new url has been added introduced (similar to create) to create
thumbs based on the THUMBNAIL_LARGE_SIZE setting
- /thumbnail/<repoid>/large/

Possible updates/tweaks for the future:
- switing to ajax method instead of image (right now the thumbnail
  generation is a blocking ajax callback)
- make cleanup urls /thumbnail/<repo>/create/(small|medium|large|xlarge)
- take care of the devices screen resolution/retina/HiDPI settings, etc.


xiez:
thx @dhoffend , really wonderful job! we will review it tomorrow.

dhoffend:
There's more improvement of course. Like the shared directory listing, or the file preview. But i've added additional comments in the issue #594 and fixing last minute types right now.

But the speed is really huge. A normal 16 MP image from recent camares is around 3-5mb while downscaled to 1280x1280 it's around 150-250kb. That makes its easy for sharing images.

static/scripts/app/views/dir.js
```diff
@@ -50,6 +51,45 @@ define([
                         tNext: gettext("Next (Right arrow key)"), // Alt text on right arrow
                         tCounter: gettext("%curr% of %total%") // Markup for "1 of 7" counter
                     },
+                    callbacks: {
+                        elementParse: function(item) {
+                            var repo_id = _this.dir.repo_id,
+                                cur_path = _this.dir.path,
+                                cur_img = _this.dir.find(function(dirent) {
+                                    return dirent.get('obj_name') == item.el[0].innerText;
```
dhoffend: Is this way okay to retrieve the file entity from the current directory? Just by having the <a> element I couldn't get a better way as using the filename as seach parameter. A better way would be to store the obj_name as data-filename attribute in the html code during directory generation.

Okay? or do you've a better solution.

xiez: @llj it looks okay to me :-)

dhoffend: As long as nobody comes up with the idea to trim the filename for better viewing experience :-)

dhoffend: I think I'll take a look into it again and add the obj_id as data-attribute to the <a> element. this way we don't run into trouble with character sets and other filename issues. Is likely the cleaner way.

dhoffend:
Added the image lightbox/gallery to the shared directory view also removed this floating preview image which gets obsolete with the new lightbox.

Since the external view seems to be oldstyle html code instead of backbone I manually copied the magnific popup code over from static to media. Hope that was right.

dhoffend:
Todo: replace the full scale image on file view (backone and html app) with the large thumbnail version to better/faster viewing. The user still can download the fullsize version

dhoffend:
I'll rename all the stuff from POPUP_DEFAULT_SIZE to THUMBNAIL_LARGE_SIZE, etc. sounds and looks better

xiez:
Hi @dhoffend , one small issue is that we already have css & js files for magnific-popup in static/, there is no need to add them again in `view_shared_dir.html`, please replace 
`<link rel="stylesheet" type="text/css" href="{{ MEDIA_URL }}css/magnific-popup.css?t=1431984103" />` with 
`<link rel="stylesheet" type="text/css" href="{% static "css/magnific-popup.css" %}" />`, and replace 
`<script type="text/javascript" src="{{ MEDIA_URL }}js/jquery.magnific-popup.js?t=1431984103"></script>` with 
`<script src="{% static "scripts/lib/jquery.magnific-popup.js" %}" />`, also plz add `{% load staticfiles %}` at top.

Thx again!

dhoffend:
@xiez thanks for pointing it out. I wasn't aware on the internal methods apart from the fact that I've never created a django or backbone application yet. Applied your fixes.

Now I also modified also the file/image preview to make use of the large thumbnail (and added the lightbox also, for better viewing experience only). This applies to internal file preview, shared single images and images for shared directories.

I think I got most image things covered. The only thing which is left is a garbage collection for invalid or unused images. This should be enabled by default if ENABLE_THUMBNAIL is set True. But this should be covered in a separate ticket.

Test builds are looking good. I'm looking forward to the review and the next point release.

xiez:
Really appreciate! Works quite well on our test server, will take a few days more to test and review. Thanks again!

dhoffend:
Hello Seafile Team. Did you had some time to review this pull request within the past month or is it already off the table to do rewrite the whole thumbnail generation stuff? I've seen changes to the view_shared_dir.html file so I guess we've to review/update my changes again.

xiez:
@freeplant

freeplant:
These changes have some edge cases not handled correctly. We are working on this feature already.

ERamseth:
Did this ever get merged?

freeplant:
We are adding grid view. This feature is not quite necessary then. On the other hand, generating and storing large thumbnails will cause storage problem at the server side.

dhoffend:
You can close the pull request or keep it for discussion. That's up to you.

tbh. If seafile/seahub really wants to provide an image gallery function it's really important to provide interner browser friendly images that are optimized for display (regarding the image size and filesize). I'm not talking about multiple image sizes for different display resolutions to optimize bandwidth, etc (like the major gallery functions are doing it (dropbox, google, etc) but at least an bandwidth/size optimized version for browsing.

The gridview should be the first step which is great, but as an enduser, if I want to take a quick look at the images on the go I don't want (or even can't) to download 2-5 megabyte per image shown in the lightbox. It's sufficent to provide a compress large version of the image for browsing and offering the enduser to download the image in full size.

If the thumbnail storage isn't able to handle large number of previews in multiple sizes (icon, grid, large, etc) this is where the next work could be done to get a great image gallery experience. If the thumbnail storage doesn't support removal of old/changed/removed images some kind of garbage collection linked together with the real backend should be considered.

ERamseth:
I tend to agree with not closing this and eventually merging the request if it works. Using full size images forthe lightbox isn't the best idea. For instance, many of the images I'm hosting and sharing via seahub are 5-10 MB JPEGs. Yes, internet is getting faster everywhere, but it's still a slowdown not only due to bandwidth but also on lower powered clients like phones and tablets that take just a little bit longer to process a file that size.

Also, assuming your seafile server has a large hard drive anyway (otherwise why would you use a small-hard drive machine to run a file server), using space by thumbnails really shouldn't be an issue.

freeplant:
This feature is finally added in version 6.1.

dhoffend:
@freeplant cool thanks. I've looked a bit into the code of the updated thumbnail process. Is it correct that the thumbnail size for popup/view is hardcoded in the templates (1024) and not a configurable variable?

tbh I would prefer 2048 for the lightbox popup (which should be configureable

Linked here:
https://github.com/haiwen/seahub/commit/5c69f79688ea71b5b75cf2078c67dec0d6ce2102#diff-1b1f2a69e2f3cfb3830de9553a2c78c9R28

freeplant:
Yes, the 1024 is hard coded. In our testing, 1024 is good for most photos. We also looked at slack. It also use 1024px images.

dhoffend:
Well. Other photo galleries use mostly 2048 to provide a better viewing expierence especially for those devices with HiDPI screens.

I'm not asking for responive image resolution just to make this size configurable. 

Anyway. Thanks :-)

freeplant:
Okay, we will add an option.

dhoffend:
Cool thumbsup :-)

PR #762, opened 2015-08-07
[fileupload] enable resumable fileupload for 'dir view' & 'upload link'

llj:
'resumable' is only enabled for 'upload files',
not for 'add folder' & 'file update'


seahub/settings.py
```diff
@@ -227,6 +227,9 @@
 # enable 'upload folder' or not
 ENABLE_UPLOAD_FOLDER = False
 
+# enable resumable fileupload or not
+ENABLE_RESUMABLE_FILEUPLOAD = False
```
freeplant: ENABLE_RESUMABLE_FILEUPLOAD change to ENABLE_RESUMABLE_FILE_UPLOAD

PR #784, opened 2015-08-26
remove operation in devices.html

Patrickgao91:


shoeper:
And why is the column being left there?

freeplant:
The column is for showing action icons. The title is not necessary.

PR #849, opened 2015-10-16
add a scrollbar when sharing libraries or folders

Patrickgao91:
scroll


shoeper:
It would look better when the share dialog would still be centered (in both dimensions).

PR #869, opened 2015-10-25
Fixing syntax error in secret_key_generator

mwd:
Ran into this one while upgrading from 4.3 to 4.4 on a raspbarry pi.


mwd:
Update path was 4.3.2 -> 4.4.3

freeplant:
Hi, the secret_key_generator is actually not needed when upgrading from v4.3 to v4.4. We have updated it in version v4.4.4.

PR #911, opened 2015-11-30
add "group-members" for redesign

Patrickgao91:


static/scripts/app/views/group.js
```diff
@@ -181,6 +194,17 @@ define([
...
         },
 
+        toggleGroupMembersPopover: function() {
+            if (!this.$('.group-members-popover').length) {
+                var $groupMembersPopover = new GroupMembersView({groupView: this});
+                this.$el.append($groupMembersPopover.$el);
+                $groupMembersPopover.show();
+            } else {
+                this.$('.group-members-popover').remove();
+                return false;
```
freeplant: Why "return false" here?

Patrickgao91: 的确不应该有这句

PR #941, opened 2015-12-11
[group members] redesigned it

llj:


static/scripts/app/views/group.js
```diff
@@ -185,6 +188,19 @@ define([
             this.hideRepoList();
             this.dirView.hide();
             this.$emptyTip.hide();
+        },
+
+        showMembers: function () {
```
freeplant: Function () 这里多了空格

PR #945, opened 2015-12-15
Search user

imwhatiam:


seahub/settings.py
```diff
@@ -227,6 +227,9 @@
 # allow seafile amdin view user's repo
 ENABLE_SYS_ADMIN_VIEW_REPO = False
 
+# allow seafile amdin view user's repo
+ENABLE_SEARCH_LDAP_USERS = False
```
freeplant: Change to ENABLE_SEARCH_FROM_LDAP_DIRECTLY

Change the description to "allow search from LDAP directly during auto-completion (not only search imported users)"

PR #956, opened 2015-12-22
show broken repo at my owned repo list page

imwhatiam:


seahub/templates/js/templates.html
```diff
@@ -36,7 +36,11 @@
       <img src="{{ MEDIA_URL }}img/sync-folder-20.png" title="{% trans "Read-Write" %}" alt="" />
       <% } %>
     </td>
+    <% if (name) { %>
     <td><a href="#my-libs/lib/<%= id %>"><%- name %></a></td>
+    <% } else { %>
+    <td>(Broken)</td>
```
shoeper: Why does it only says (Broken) without any information? Wouldn't it still be helpful to put the id in? Maybe with a small info icon next to it which says something like `please contact your administrator providing the library id to fix this library`?

freeplant:
@shoeper Showing library id is not necessary. The admin can search the user then check the library list.

PR #957, opened 2015-12-22
Make user search case insensitive

egroeper:
If you want to find "Carl Smith", typing "carl" should list him, too.
People are lazy and expect case insensitive searches.
They don't think "Oh. I have to type 'Carl' instead.". They think
"Search doesn't work.".


freeplant:
Thank you very much. Can you send a pull request to 5.0 branch instead? I'd like to include it in version 5.0.

egroeper:
Obsoleted by #958.

PR #958, opened 2015-12-22
Make user search case insensitive

egroeper:
If you want to find "Carl Smith", typing "carl" should list him, too.
People are lazy and expect case insensitive searches.
They don't think "Oh. I have to type 'Carl' instead.". They think
"Search doesn't work.".

### DIFF
--- a/seahub/api2/views.py
+++ b/seahub/api2/views.py
@@ -313,12 +313,12 @@ class SearchUser(APIView):
 
                 searched_users = filter(lambda u: q in u.email, users)
                 # 'user__in' for only get profile of user in org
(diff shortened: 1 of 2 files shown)


freeplant:
It is done in https://github.com/haiwen/seahub/commit/a722448548d08eb8d39a3288cc39ce66604102ff . We have to modify a few lines in test to make it work in 5.0 branch.

egroeper:
I'm fine with that. But you should fix your test (I made a line comment in the commit).

freeplant:
Fixed in https://github.com/haiwen/seahub/pull/982/files

PR #975, opened 2015-12-31
Fix for ampersand in file path of download link

ERamseth:
Without this, the big green download link will not work when there is an ampersand in the file path.


llj:
Thanks for the fix.
I have put it on branch 5.0:
https://github.com/haiwen/seahub/commit/5a167488e7051d18683155a57891c251fbf499af

PR #1028, opened 2016-02-23
[shib] Use display_name in shib_meta as nickname, fallback to givenna…

xiez:
…me + surname if missing


egroeper:
As requested my review:
Looks fine to me. For sake of completeness I would have added an additional test, where surname, givenname and display_name are given. That way you could test, if display_name really gets preference.

Thanks for implementing my feature request!

xiez:
@egroeper  tests updated, thanks!

in https://forum.edu.seafile.de/t/displayname-as-shibboleth-attribute/89, u said update account api didn't refresh nickname cache, but that bug should be fixed in version 4.4.7 pro, can u reproduce that in version 5.0.x ?

PR #1099, opened 2016-04-03
Support contact_email in email sending

egroeper:
support contact_email as from and in Reply-to header.

Apart from that I improved the share-link tests in b64e882.


xiez:
@egroeper  really nice work! Thanks a lot!

PR #1104, opened 2016-04-06
API to send a email notification about a conflict in the client files.

mrakob:
Notification will be sent from seafile-client

Changes to be committed:
    new file:   seahub/api2/endpoints/send_notify_file_conflict.py
    modified:   seahub/api2/urls.py
    modified:   seahub/settings.py
    new file:   seahub/templates/send_email_notify_file_conflict.html


freeplant:
Hi mrakob, what's the usage case of this API?

shoeper:
@freeplant it should allow the client to send an email (to the user) in case there is a conflict.

freeplant:
@mrakob do you call this API from desktop client? Have you changed the desktop client?

seahub/api2/endpoints/send_notify_file_conflict.py
```diff
@@ -0,0 +1,79 @@
...
+        path = request.data.get('path', '')
+
+        from_email = None  # use default from email
+
+        title = _(u'File conflict notification %s') % SITE_NAME
+        template = 'send_email_notify_file_conflict.html'
+
+        to_email = email
+
+        logger.warning('repo_id: %s repo_name: %s, path: %s', repo_id, repo_name, path)
```
shoeper: What should this be good for?

shoeper:
I like the idea but I would default the setting (IS_SEND_EMAIL_FILE_CONFLICT = True) to false. It should be an active decision for the administrator because it can lead to many emails (thousands of emails).

freeplant:
I think the idea of letting the client call the server to send a conflict notification email is not good. If it is really need, it is better to let the server detects such conflicts and notify users.

PR #1166, opened 2016-04-29
send file audit msg & update quota when view raw shared file

imwhatiam:


seahub/views/file.py
```diff
@@ -779,6 +779,15 @@ def view_shared_file(request, fileshare):
                                                            use_onetime=False)
     raw_path = gen_file_get_url(access_token, filename)
     if request.GET.get('raw', '') == '1':
+        # check whether owner's traffic over the limit
+        if user_traffic_over_limit(shared_by):
+            messages.error(request, _(u'Unable to view raw file, share link traffic is used up.'))
+            return HttpResponseRedirect(next)
```
lins05: Where is `next` defined?

PR #1229, opened 2016-06-10
Update django.po

Mykos:
Typo correction.


freeplant:
Please modify the translation on Transfix. More information can be found here: https://github.com/haiwen/seafile/wiki/Seahub-Translation#submit-translation

PR #1291, opened 2016-07-19
[wiki] modification

llj:


shoeper:
It would be very helpful if you could elaborate a bit in your commit messages to allow others to see what has been changed. Something like `[wiki] modification` can be almost everything.

PR #1388, opened 2016-11-02
check if user num exceed limit before add user

imwhatiam:


seahub/utils/__init__.py
```diff
@@ -19,7 +19,7 @@
 import ccnet
 from constance import config
 import seaserv
-from seaserv import seafile_api
+from seaserv import seafile_api, ccnet_api
```
xiez: 需要删除

seahub/utils/licenseparse.py
```diff
@@ -36,3 +46,23 @@ def parse_license(file_path):
...
+
+            # get active user number
+            active_db_users = ccnet_api.count_emailusers('DB')
+            active_ldap_users = ccnet_api.count_emailusers('LDAP')
+
+            return active_db_users + active_ldap_users >= max_users
+        except Exception as e:
+            logger = logging.getLogger(__name__)
+            logger.error(e)
+            return False
```
xiez: return True

seahub/utils/licenseparse.py
```diff
@@ -36,3 +46,23 @@ def parse_license(file_path):
             ret[k.strip()] = v.strip().strip('"')
 
     return ret
+
+def user_number_over_limit():
```
xiez: def user_number_over_limit(extra_num=0):

PR #1432, opened 2016-12-23
Add missing translations for Sysadmin settings page

jamietanna:


freeplant:
This variables don't need to be translated. They should be consistent with the variable names in Seafile manual.

PR #1433, opened 2016-12-23
Add instructions for how to set up in a virtualenv

jamietanna:
Using a virtualenv is much better than globally installing all of the
project's dependencies, as it doesn't pollute the global package space,
and allows us to work just within the project folder.


freeplant:
Thank you.

PR #1440, opened 2017-01-06
Use seafile_api to set/get org user quota

cuihaikuo:


seahub/views/ajax.py
```diff
@@ -1355,8 +1355,7 @@ def space_and_traffic(request):
         space_usage = seafile_api.get_user_self_usage(username)
     else:
         org_id = org[0].org_id
-        space_quota = seafserv_threaded_rpc.get_org_user_quota(org_id,
-                                                               username)
+        space_quota = seafile_api.get_org_user_quota(org_id, username)
         space_usage = seafserv_threaded_rpc.get_org_user_quota_usage(
```
freeplant: This line `seafserv_threaded_rpc.get_org_user_quota_usage` need to be changed too.

freeplant: This line `seafserv_threaded_rpc.get_org_user_quota_usage` need to be changed too.

PR #1582, opened 2017-05-09
[onlyoffice] Fix document key

blaniel:
According to https://api.onlyoffice.com/editors/config/document, the key
should be changed when document is edited *and* saved. Using a different
key each time the document is opened tells document server a different version
of the file is edited and thus breaks save operations as well as collaborative
editing.

Use obj_id as key since it's a file content based uuid.


xiez:
Okay, thanks! @blaniel

PR #1596, opened 2017-05-19
add option to set language of OnlyOffice DocumentServer

TMHBOFH:
With the parameter ONLYOFFICE_EDITOR_LANGUAGE_SETTING  in seahub_settings.py you can set hard the language from OnlyOffice DocumentServer.
If the parameter not set, will the language dynamice determined from client language.


xiez:
Thanks @TMHBOFH , we will review your work in these days.

xiez:
Merged! Many thanks!

PR #1617, opened 2017-06-04
Possibility to Support OpenDocument File by OnlyOffice

TMHBOFH:
Hi @xiez
the following PR make it possible to edit OpenDocument files (odt,ods,odp) with OnlyOffice.
A few words before that. ONLYOFFICE Document Server use the Office Open XML formats for editing files. That meens for my solution if you want editing a odt file becomes this twice converting.

_seafile -[odt]-> ONLYOFFICE Document Server {editing} -[docx]-> seafile -[docx]-> OO document conversion service -[odt]-> seafile save the file_

Must everyone know if the two-time converting ok is???

I ask for forbearance i'm not a program. (I know, the error error handling and so on is not so good)

parameter description:

ONLYOFFICE_EDIT_FILE_EXTENSION = ('odt', 'ods', 'odp', 'docx', 'pptx', 'xlsx')
types of files should be editable through OnlyOffice

ONLYOFFICE_CONVERTSERVICE_URL = 'https://[OnlyOfficeDocuemntServer]/ConvertService.ashx'
url of document conversion service  page

I would be nice if you could look at this if you have time.


freeplant:
Is this two-time converting reliable? I'm afraid after using this feature, the format of the original OpenDocument files will not be well-preserved.

DerDanilo:
The OnlyOffice Conversion service could be of help here. I mean they support to edit those files. 

Maybe one should note that when editing .odt etc. with OnlyOffice the files will be saved in .docx .xlsx etc. afterwards. That solution would clearly work for me, as long as the format is kept somewhat correct. To what I have seen the OnlyOffice Editor works absolutely fine with formating the files.

TMHBOFH:
@freeplant
i have the same apprehension with two-time converting, but at all my test it works fine.
My suggestion would be, you take up this function with the reference in the documentation on the two-time converting.(@DerDanilo maybe youre part)

Important would be to me, that you check and improve the source code.

DerDanilo:
I can add it to the documentation once it's included into the code and published as stable. Don't want to recommend something that is not stable yet. :)

alexanderonlyoffice:
Hello @DerDanilo, we have noticed that the solution was not added to the release. We have some notes that can be helpful.

Please note that ONLYOFFICE works with Office Open XML format files only and this format is considered as base format.
Other formats are converted before opening. File is given back in base format when saving.
It is incorrect to open the file of .odt format for editing in ONLYOFFICE and to leave the format as it is for viewing in document manager. 
File gets the formatting and structure of .DOCX after saving, but is shown as .ODT. If one downloads the file and tries to open it in OpenOffice or LibreOffice, he/she will get an error.
We can offer several options of the solution:
* To convert the Open Document Format file into Office Open XML before opening and offer the user to edit the new file;
* To change the extension of Open Document Format file after the editing to the corresponding Office Open XML format;
* To convert the received Office Open XML file after the editing of Open Document Format back to Open Document Format when saving the file.
Please note that distortion of the content is possible while converting due to the differences between the formats.
Our API documentation for the conversion is [here](https://api.onlyoffice.com/editors/conversionapi).

PR #1641, opened 2017-06-15
[preview-psd]  preview psd file in server

zMingGit:


seahub/views/file.py
```diff
@@ -310,6 +310,13 @@ def file_size_exceeds_preview_limit(file_size, file_type):
             return False, ''
     elif file_type in (VIDEO, AUDIO):
             return False, ''
+    elif file_type == PSD:
+        if file_size > 200000000:
```
shoeper: Please move it to a setting

seahub/thumbnail/utils.py
```diff
@@ -6,6 +6,8 @@
 import urllib2
 import logging
 from StringIO import StringIO
+from PIL import Image, ImageCms
+from psd_tools import PSDImage
```
shoeper: Looks like it adds a dependency. Needs documentation

seahub/views/file.py
zMingGit: ok

seahub/thumbnail/utils.py
zMingGit: sry, i can ask if it is about psd tools or just what I wrote about psd_tools code?

shoeper: It is about psd_tools. Is it bundled or installed by default? Otherwise I think this needs documentation how to set it up.

PR #1688, opened 2017-07-11
check if is org user before share repo/dir to user

imwhatiam:


seahub/api2/endpoints/dir_shared_items.py
```diff
@@ -298,6 +300,18 @@ def put(self, request, repo_id, format=None):
                 try:
                     if is_org_context(request):
                         org_id = request.user.org.org_id
+
+                        if not is_org_user(to_user, int(org_id)):
+                            org_name = request.user.org.org_name
+                            error_msg = _(u'User %s is not member of organization %s.') \
```
xiez: 这个不需要翻译了

seahub/api2/endpoints/dir_shared_items.py
```diff
@@ -306,6 +320,15 @@ def put(self, request, repo_id, format=None):
                             sub_repo_id = seafile_api.org_share_subdir_to_user(org_id,
                                     repo_id, path, username, to_user, permission)
                     else:
+
+                        if is_org_user(to_user):
```
xiez: org id

PR #1692, opened 2017-07-12
[system admin] traffic: add tip

llj:


seahub/templates/sysadmin/sys_trafficadmin.html
```diff
@@ -7,6 +7,8 @@
 {% block right_panel %}
 <h3 class="hd">{% trans "User Traffic" %}</h3>
 
+<p class="tip">{% trans "Tips: the traffic here only includes the traffic used by sharing links." %}</p>
```
shoeper: Tip: the traffic only includes the traffic used by sharing links.

freeplant: Thanks.

PR #1712, opened 2017-07-20
[repo snapshot labels] added 'label current state' for my libs; show …

llj:
…labels in 'repo history' page


seahub/views/__init__.py
```diff
@@ -538,6 +539,17 @@ def repo_history(request, repo_id):
     for c in commits:
         c.show = False if new_merge_with_no_conflict(c) else True
 
+    show_label = False
+    if ENABLE_REPO_SNAPSHOT_LABEL and user_perm == 'rw':
```
freeplant: 'rw' -> 'r'

PR #1715, opened 2017-07-21
rm authentication check when getting uploaded bytes

imwhatiam:


shoeper:
Is this to fix upload links? How is being made sure that I don't upload to another library or path?

freeplant:
It is only for checking the uploaded bytes (how many bytes are uploaded already for a specific file) for doing resumable upload.

shoeper:
Thanks.

PR #1729, opened 2017-07-28
set default role permission

imwhatiam:


xiez:
no need for this PR, ref: https://github.com/haiwen/seahub/commit/190bf7d66b256e5cf7b9c2ea79a4dad892d3a96e

PR #1805, opened 2017-09-11
search repo by user

zMingGit:


seahub/api2/views.py
```diff
@@ -536,6 +538,46 @@ def get(self, request, format=None):
...
+            owned_repos.sort(lambda x, y: cmp(y.last_modify, x.last_modify))
+            for r in owned_repos:
+                # do not return virtual repos
+                if r.is_virtual:
+                    continue
+
+                shouldIn = False
+                if not q:
+                    shouldIn = True
+                if q and q.lower() in r.name.lower():
```
freeplant: if q and q.lower() not in r.name.lower():
    continue

PR #1849, opened 2017-10-04
seahub.invitations: Give more context to template

moschlar:
So that e.g. the accepter can be addressed more directly or different template texts can be used depending on the accepter's domain.


xiez:
@moschlar  Thanks! merged.

PR #1852, opened 2017-10-09
is_dir always is true

zMingGit:


seahub/api2/endpoints/file_tag.py
```diff
@@ -86,7 +86,7 @@ def _decorated(view, request, *args, **kwargs):
```
xiez: bool 参数应该在检查时转成python 里的true，false。参照 https://github.com/haiwen/seahub/blob/master/seahub/api2/endpoints/admin/users.py#L180

PR #2003, opened 2018-02-02
add api for upload file via upload link foken

imwhatiam:


seahub/api2/endpoints/upload_links.py
```diff
@@ -220,3 +221,49 @@ def delete(self, request, token):
             return api_error(status.HTTP_500_INTERNAL_SERVER_ERROR, error_msg)
 
         return Response({'success': True})
+
+
+class UploadLinkUpload(APIView):
+
+    throttle_classes = (UserRateThrottle, )
```
xiez: 应该用anon throttle

PR #2046, opened 2018-03-21
forcesave onlyoffice

salleman33:
save the document when onlyoffice performs a forcesave request


imwhatiam:
Hello, the OnlyOffice docs says that if status is 6, it means that `forcesave` this document, as it is not the final version, so we didn't deal with this logic.

```
force saving the document being edited without closing it (the document editing might be continued after this command, so this will not be the final saved document version);
```

image

imwhatiam:
Or is there something we haven’t considered and need to do this?

freeplant:
The feature is included in the latest release.

PR #2057, opened 2018-03-30
make the change password optional

ilearnit:


seahub/settings.py
```diff
@@ -323,6 +323,9 @@
 # Force user to change password when admin add/reset a user.
 FORCE_PASSWORD_CHANGE = True
 
+# Make the change password optional
```
llj: Enable a user to change password in 'settings' page.

PR #2074, opened 2018-04-18
Delete user log after deleting user

ilearnit:


xiez:
ref: https://github.com/haiwen/seahub/commit/acd4c2456de96ce3ba93d179a24086289a8a7a0e

PR #2092, opened 2018-04-28
filter out address book gorup when:

imwhatiam:
1. share library/folder
2. set folder permission


llj:
`media/assets/` 下的文件是不用改的。

PR #2109, opened 2018-05-09
[address book] added check for invalid input when set group quota

llj:


shoeper:
Isn't 0 == unlimited?

Issue #2164, opened 2018-06-16
Web UI will not autoload a >100 item directory view 


rallfo:
### Description
In certain circumstances, on multiple platforms & browsers (Firefox + Chrome) incl mobile the Web UI will fail to correctly load the next 100 items in large directories.

### Reports
#2085 Web UI in mobile browser does not auto load
[Seafile Forums - October 2017 - Only first 100 folders being displayed in Firefox](https://forum.seafile.com/t/only-first-100-folders-being-displayed-in-firefox/4263)

### Replicating
Chrome 67 on Windows 10 @ 1440p resolution with page zoomed into 110: large directories fail to load past 100 items. When zoomed to 100% or 125%: large directories load correctly.
See reports for further examples.

### Cause
Flawed check in Javascript method responsible for detecting the bottom of page condition failing with  sub-pixel coordinates:

In [_static/scripts/app/views/dir.js_](https://github.com/haiwen/seahub/blob/32019fe3054585175b82735360371927295c59fe/static/scripts/app/views/dir.js) this is the check for scrolling to the bottom of the page and calling **render_dirents_slice()** to render the next 100 items:

```javascript
onWindowScroll: function () {
                // 'more'
                if (this.dir.dirent_more &&
                        $(window).scrollTop() + $(window).height() == $(document).height()) { // scroll to the bottom
                    this.render_dirents_slice(this.dir.last_start, this.dir.limit);
                    this.getThumbnail();
                }
```

However this check is flawed:
- scrollTop et al are actually floating point values in browsers with sub-pixel coordinates and are fractional on high DPI displays, when zoomed in and other circumstances. Floating point numbers should never be compared with == . See this [chromium bug #489618](https://bugs.chromium.org/p/chromium/issues/detail?id=489618)
- Generally better UI behaviour is to trigger within a given distance of the bottom rather than the exact last pixel, especially for scrolling on mobile. i.e. within 25% of the window client height

rallfo:
Fix for **6.2** in PR #2165 

6.3 appears to have the same bug needs testing within new design (I am not running it)

PR #2165, opened 2018-06-16
BUGFIX: Web UI will not autoload a >100 item directory view #2164

rallfo:
### Please Refer to Issue #2164 
_Web UI will not autoload a >100 item directory view_

**Fix for 6.2, 6.3 appears to have the same bug but needs separate patch**


freeplant:
It has been fixed in https://github.com/haiwen/seahub/commit/8e9636a3a0f1217de575aeb87489529d3cbde85b

PR #2189, opened 2018-06-30
Set autofocus in two factor form with single input field

rene-s:
This enables users to enter the OTP token right after login without manually focusing the single input field.


schneimi:
+1

freeplant:
Thank you

PR #2279, opened 2018-08-16
update admin libraries trash page

imwhatiam:
show expire days according to seafile.conf configuration


seahub/views/sysadmin.py
```diff
@@ -97,6 +97,12 @@ def sysadmin(request):
 
     folder_perm_enabled = True if is_pro_version() and settings.ENABLE_FOLDER_PERM else False
 
+    try:
+        expire_days = seafile_api.get_server_config_int('library_trash', 'expire_days')
+    except Exception as e:
+        logger.error(e)
+        expire_days = -1
```
xiez: expire_days = 30

下面就不用判断 > 0 了

PR #2285, opened 2018-08-18
Token time

guopengda:


seahub/api2/authentication.py
```diff
@@ -71,6 +74,11 @@ def authenticate_v1(self, request, key):
         except Token.DoesNotExist:
             raise AuthenticationFailed('Invalid token')
 
+        # expired token API_TOKEN_AGE
+        if API_TOKEN_AGE == 7 * 24 * 60 * 60:
```
shoeper: This looks wrong to me. It'll only check expiry for tokens having the currently set API_TOKEN_AGE, but there can be tokens with a different token age and those will be valid forever.

PR #2292, opened 2018-08-21
New token time

guopengda:


omkarnathsingh:
What is benefit of this feature?

shoeper:
@omkarnathsingh so far tokens never expire. When enabling this api token will expire after a given time.

omkarnathsingh:
Ah, nice. Let me try it and add to my seafile. Thanks

Sent from ProtonMail mobile

-------- Original Message --------

shoeper:
Main question I'd have it this also effects the clients. There one usually doesn't want to get logged out.

omkarnathsingh:
@shoeper Can we have better have it as a setting for SYstem admin to either enable or not. I feel if its added, it will expire token for android/windows app also making it to login again and again.

PR #2320, opened 2018-08-30
Remove unreachable code

bougault:
```python
    if getattr(settings, 'ENABLE_OAUTH', False):
        return HttpResponseRedirect(reverse('oauth_login') + next_param)

    if getattr(settings, 'ENABLE_CAS', False):
        return HttpResponseRedirect(reverse('cas_ng_login') + next_param)

    if getattr(settings, 'ENABLE_OAUTH', False):
        return HttpResponseRedirect(reverse('oauth_login'))
```
The 'ENABLE_OAUTH block is duplicated, so the second one is dead code.


shoeper:
Are you sure it is unreachable? It depends on the configuration.

bougault:
```python
    if getattr(settings, 'ENABLE_OAUTH', False):
        return HttpResponseRedirect(reverse('oauth_login') + next_param)

    if getattr(settings, 'ENABLE_CAS', False):
        return HttpResponseRedirect(reverse('cas_ng_login') + next_param)

    if getattr(settings, 'ENABLE_OAUTH', False):
        return HttpResponseRedirect(reverse('oauth_login'))
```
Yes, in this snippet, third condition is the same as the first one so it can be reached

shoeper:
Sorry, totally missed that.

xiez:
Merged! Thanks guys!

PR #2328, opened 2018-09-01
update share link permission

imwhatiam:
add can_preview field for file when ajax list lib dir
hide preview-only permission if a file can not preview

### DIFF
--- a/seahub/templates/js/templates.html
+++ b/seahub/templates/js/templates.html
@@ -1162,7 +1162,7 @@
     <div id="share-tabs" class="nav-con-tabs">
         <ul class="nav-con-tabs-nav">
             <% if (!repo_encrypted && can_generate_share_link) { %>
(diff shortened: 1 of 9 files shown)


seahub/templates/js/templates.html
```diff
@@ -1162,7 +1162,7 @@ <h3 class="hd" id="dialogTitle"><%= title %></h3>
     <div id="share-tabs" class="nav-con-tabs">
         <ul class="nav-con-tabs-nav">
             <% if (!repo_encrypted && can_generate_share_link) { %>
-            <li class="tab"><a href="#download-link-share" class="a">{% trans "Download Link" %}</a></li>
+            <li class="tab"><a href="#download-link-share" class="a">{% trans "Share Links" %}</a></li>
```
freeplant: 改成 Share Link

PR #2336, opened 2018-09-04
[api] Add draft feature

xiez:


xiez:
ref: https://github.com/haiwen/seahub/pull/2375

PR #2554, opened 2018-11-22
Add webdav secret api

xiez:


seahub/options/models.py
```diff
@@ -28,6 +28,7 @@
 VAL_USER_LOGGED_IN = "1"
 
 KEY_DEFAULT_REPO = "default_repo"
+KEY_WEBDAV_SECRET = "webdav_secret"
```
shoeper: It would be better to not have such a weak default, but let the code fail when no secret was configured instead.

xiez: It's just a user option type, nothing to do with any security secret.  The name may seems misleading through.

PR #2595, opened 2018-12-04
wip: psd preview

xiez:


JIMhackKING:
Now I want to implement the preview PSD file function in version 6.2 myself, but is this required to be supported by the seafile code server? The network always request the seafileserver/files/ address , and return the 400 code to me (Response body is "Bad access token")

PR #2641, opened 2018-12-13
fix-selection

Michael18811380328:
1. Make sure that range is right and common after we clicked add-comment button.
2. Remove null nodes in fragment if we select several paragraphs.


frontend/src/draft-review.js
```diff
@@ -290,6 +297,27 @@ class DraftReview extends React.Component {
     this.oldIndex = node.data.get('old_index');
   }
 
+  removeNullNode = (oldNodes) => {
+    let newNodes = [];
+    oldNodes.map((node) => {
+      const text = node.text.trim();
```
freeplant: if (!text) {
  return;
}

PR #2795, opened 2019-01-09
import-group-members

Michael18811380328:
Import group members by uploading a cvs file.


Michael18811380328:
This function is unnecessary, so we don't use it now. And there are more important things we should deal with.

PR #2848, opened 2019-01-17
Fixed folder share notification

rabb1t:
Fix folder share notification -- Folder name must be specified inside html's `<a href=` block

### DIFF
--- a/seahub/notifications/templates/notifications/notice_msg/folder_share_msg.html
+++ b/seahub/notifications/templates/notifications/notice_msg/folder_share_msg.html
@@ -1,1 +1,1 @@
-{% load i18n %}{% blocktrans %}{{user}} has shared a folder named <a href="{{url_base}}{{lib_url}}">{{lib_name}}</a> to you.{% endblocktrans %}
+{% load i18n %}{% blocktrans %}{{user}} has shared a folder named <a href="{{url_base}}{{lib_url}}{{lib_name}}">{{lib_name}}</a> to you.{% endblocktrans %}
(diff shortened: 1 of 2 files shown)


freeplant:
Hi,

The code in master release is not released yet. How do you find the problem?

rabb1t:
Hi, the notification template is the same in `master`, `seafile-pro-server-6.3.11` and versions below. I've just create a pull-request against `master` branch where this issue still persists.

xiez:
Hi @rabb1t , can you provide steps to reproduce this issue/bug ? and some screenshots would be helpful.

As we tested in version 6.3.x, the href link without folder name works fine. It would be broken if your patch is applied.

rabb1t:
Sure, here are the steps to reproduce:

1) Inside the seafile library create new folder
2) Then select "Share"->"Share to user" option to share this folder
3) Now user which you have specified in Step 1 should receive notification message with the URL which looks like this:
`https://your-domain.example.com/#common/lib/e147f832-2d89-401c-9077-2fa8433fc232/`

   which navigates to the library instead of shared folder.
4) After clicking that link you will see the "Permission denied." error:

freeplant:
Hi,

We can't reproduce the problem. For step 3:

image

The be-shared user will see the above message in the notification list. After click the link, he will see:


image


Note, the library ID in the URL is `0153f6de-xxxx`.  This is a virtual library ID for the sub-folder.

The origin library ID is `8342175d-xxxx`:



image

----

Can you check the original library and the library ID for the virtual library for the be-shared sub-folder in your case?

rabb1t:
> The be-shared user will see the above message in the notification list. After click the link, he will see:

Sorry, I forgot to mention that I am talking about email notification template, not the browser notices. Did you receive email with the shared link?

xiez:
Hi @rabb1t , this bug is fixed in https://github.com/haiwen/seahub/pull/3134/files, which will be included in v6.3.13 pro. Thanks!

PR #2925, opened 2019-02-08
Add proxy authentification for simple SSO

artonge:
In some cases, applications can trust the proxy to pass HTTP headers to authenticate a user. This allow a simple form of SSO.

I  added this possibility in seahub.

When the `TRUST_PROXY_AUTHTENTICATION` variable is set to `True`, seafile will create a session for user passed in the `Remote-User` HTTP header. The used HTTP header can be modified by setting the `PROXY_AUTH_HEADER` variable.

The code come mostly from the [YunoHost Seafile package](https://github.com/YunoHost-Apps/seafile_ynh/blob/master/sources/patches/x86-64-sso_auth.patch) written by @Josue-T.

As I never used Django, I am not sure that I placed the code at the correct places. I will be happy to make some corrections if needed.

### DIFF
--- a/seahub/auth/middleware.py
+++ b/seahub/auth/middleware.py
@@ -1,7 +1,8 @@
 # Copyright (c) 2012-2016 Seafile Ltd.
-from django.contrib import auth
+from seahub import auth
 from django.core.exceptions import ImproperlyConfigured
 
+from django.conf import settings
 
 class LazyUser(object):
     def __get__(self, request, obj_type=None):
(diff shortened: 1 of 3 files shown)


xiez:
Thank you! @artonge 
We'll review this PR in next week.

xiez:
Hi, this PR is merged into master which will be included in version 7.0
BTW,  
I added some code to avoid potential security issue, see https://github.com/haiwen/seahub/commit/b3e950209d6c7d0a755644fed7bc3b0c7bd529a6

Thanks again!

PR #3159, opened 2019-03-22
check heading increase

WangJianhui666:


WangJianhui666:
[Heading levels should only increment by one level at a time](https://trello.com/c/9PWnR3Ll/655-heading-levels-should-only-increment-by-one-level-at-a-time)

PR #3334, opened 2019-04-19
fix warning in other format of files

LeoSirius:


frontend/src/components/file-view/file-toolbar.js
```diff
@@ -10,6 +10,9 @@ import ShareDialog from '../dialog/share-dialog';
 const propTypes = { 
   isLocked: PropTypes.bool.isRequired,
   lockedByMe: PropTypes.bool.isRequired,
+  onSaveChangedContent: PropTypes.func,
```
freeplant: onSaveChangedContent -> onSave

frontend/src/components/file-view/file-toolbar.js
```diff
@@ -10,6 +10,9 @@ import ShareDialog from '../dialog/share-dialog';
...
+  isSaving: PropTypes.bool,
+  isContentChangedButNotSaved: PropTypes.bool,
```
freeplant: needSave

Issue #2957, opened 2019-02-18
Bug with ForceSave and OnlyOffice


salleman33:
Hi, 

I have a problem with the "forceSave" functionnality of OnlyOffice.
my test is : 
- User A open a file in OnlyOffice, makes some change and use the forceSave functionnality
- User B open the same document
result :  UserB is not on the same instance of OnlyOffice document that User A

I think the problem comes from the "doc_key" id. OnlyOffice API documentation says  : 

> Defines the unique document identifier used for document recognition by the service. In case the known key is sent the document will be taken from the cache. Every time the document is edited and saved, the key must be generated anew. The document url can be used as the key but without the special characters and the length is limited to 20 symbols.

I suppose that user B is on a new instance because doc_key changes. In the onlyoffice callback, the key_doc id have to not  change when the status is 6 (forceSave), but only if status is 2 (really save when all users leave onlyoffice document)

Regards

freeplant:
This is a known problem. It is a restriction of OnlyOffice. I suggest you to not use 
forceSave option.

odontomachus:
Sorry to comment on a closed issue. How about using a fixed doc key while there is a collaborative editing session?

https://github.com/odontomachus/seahub/tree/feature_onlyoffice_sessions

Issue #1854, opened 2017-10-10
"Last Change" Column not updated after editing with Only office


YannikG:
I'm wondering why doesn't Seafile generate a new Version of a File when I'm editing this File in the Only Office Document Online Document Editor?  Seafile does also not update the "Last Change" Column in the View.

This Bug cause that the SeaDrive Client does not sync the Files.

Thanks for this great Product!

Hope this Bug get fixed soon

shoeper:
Is this still the case?

YannikG:
Could be. I am not using Seafile anymore.

odontomachus:
This is due to the integration workflow.
OnlyOffice only saves the document to the backend when **all editing sessions are closed**. There is the same problem with NextCloud & OwnCloud's integrations. The temporary solution we have come up with for our project was to shorten the OnlyOffice editor session so that it "kicks" everyone out after a period of inactivity and saves the document. We found a session of about 10 minutes to work reasonably well, so long as users are aware of this limitation.

See: https://github.com/ONLYOFFICE/onlyoffice-owncloud/issues/116

Currently, enabling the force save option in OnlyOffice breaks the Seafile integration model for collaborative editing. Saving changes the version of the file, and that is used to build the OnlyOffice session url, so if a second user joins in after the first user has saved, she will be working on a different version of the document in a separate session.

I am working on an alternative model, but need to figure out the edge cases. Once that's done, I'll make a PR.
https://github.com/odontomachus/seahub/tree/feature_onlyoffice_path_based

PR #3604, opened 2019-06-03
Use a running onlyoffice doc_key during a collaborative editing session.

odontomachus:
Use a running OnlyOffice session for concurrent edits of a file.

When a user opens an OnlyOffice document for editing without specifying a version, the session key is cached. Any subsequent user opening the document before the end of the concurrent editing session will use the same doc_key, and join the same editing session.

Currently, if a user saves a document in OnlyOffice through force save or some timeout, the file_id in Seafile changes, and new users opening the document are sent to a different OnlyOffice editing sesison. This leads to forked versions fo the file and users in separate sessions.

See https://github.com/haiwen/seahub/issues/1854 and https://github.com/haiwen/seahub/issues/2957

This has been tested on Seafile 6.3.4 CE.

### DIFF
--- a/seahub/onlyoffice/utils.py
+++ b/seahub/onlyoffice/utils.py
@@ -20,6 +20,7 @@ def get_onlyoffice_dict(username, repo_id, file_path,
         file_id='', can_edit=False, can_download=True):
 
     repo = seafile_api.get_repo(repo_id)
+    latest = can_edit and (not file_id)
     if repo.is_virtual:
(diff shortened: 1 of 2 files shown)


dani:
We're facing this issue as well, forcing us to disable auto save (which can be a bug problem for big docs, when editing last for hours, we do not save intermediate versions). Any news for this PR ? Was it reviewed ?

dani:
Was this PR merged ?

PR #3607, opened 2019-06-03
build: parametrize python & django-admin.py

commodo:
This is particularly useful when cross-building Seafile in build-systems
such as OpenWrt or buildroot, where the `python` executable is not in PATH.

Manipulating the PATH variable is also doable but, it's better to provide
the full/absolute path to python, as well as to the django-admin.py script.

This change allows python & django-admin.py to be specified at a different
location, as well as maintaining their default values if unspecified.

Strictly speaking for cross-building, this change is sufficient.

Signed-off-by: Alexandru Ardelean <ardeleanalex@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,6 @@
+PYTHON ?= python
+DJANGO_ADMIN_PY ?= django-admin.py
+
 PROJECT=seahub
 
 develop: setup-git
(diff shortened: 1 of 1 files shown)


commodo:
Changelog v1 -> v2:
* rebased on a newer base branch; 

Note: the build looks unrelated to this change.

Any thoughts? Input?

commodo:
Changelog v2 -> v3:
* rebased on newer master

freeplant:
Sorry for missed the PR for a long time. We will check it.

freeplant:
The scripts is used to build JS files. The JS files are actually already built in GitHub actions and push to a branch like 7.1-dist. So the modification is not needed. If you build a Seafile from the source code, just use 7.1-dist branch.

PR #3690, opened 2019-06-21
Putting the file out is the same as putting it in a blank space

zxj96:


frontend/src/components/dirent-list-view/dirent-list-view.js
```diff
@@ -71,6 +71,7 @@ class DirentListView extends React.Component {
       isListDropTipShow: false,
     };
 
+    this.dragEnterContainerAmount = 0;
```
shuntian: dragEnterContainerAmount --> enteredCounter

PR #3697, opened 2019-06-21
repair share btn control bug

shuntian:


frontend/src/utils/utils.js
```diff
@@ -901,6 +901,50 @@ export const Utils = {
     }
     return event.target.getAttribute('data-' + data);
     
+  },
+
+  /**
+   * judgement user is has the permission to share.
```
freeplant: Check whether user has permission to share

frontend/src/utils/utils.js
```diff
@@ -901,6 +901,50 @@ export const Utils = {
...
+   * one scence: current repo(the root path), in the toolbar
```
freeplant: scene 1:

frontend/src/utils/utils.js
```diff
@@ -901,6 +901,50 @@ export const Utils = {
...
+   * two scence: selected a dirent, in the toolbar
```
freeplant: scene 2:

frontend/src/utils/utils.js
```diff
@@ -901,6 +901,50 @@ export const Utils = {
...
+   * three scence: dirent list(grid list), in the item operation menu;
```
freeplant: scene 3:

frontend/src/utils/utils.js
freeplant: If dirent is none, then check whether the user can share the repo

PR #3974, opened 2019-08-12
sysadmin settings api

LeoSirius:


seahub/views/sysadmin.py
```diff
@@ -2235,6 +2235,7 @@ def sys_settings(request):
     config_dict = {}
     for key in dir(config):
         value = getattr(config, key)
+        print(str(key) + ': ' + str(value))
```
freeplant: Mark

seahub/api2/endpoints/admin/web_settings.py
```diff
@@ -0,0 +1,89 @@
...
+        STRING_WEB_SETTINGS = ('SERVICE_URL', 'FILE_SERVER_ROOT', 'TEXT_PREVIEW_EXT',
+                               'SITE_NAME', 'SITE_TITLE', 'CUSTOM_CSS')
+
+        res = []
+        for key, value in request.data.items():
+            if key not in dir(config) or value is None:
+                error_msg = _(u'setting invalid.')
+                return api_error(status.HTTP_400_BAD_REQUEST, error_msg)
+
+            if value.isdigit():
```
imwhatiam: if key not in str_s and not in digit_s:
   error


if key in str_setting and value not is str:
    error

if key in digit_setting and value is not digit:
    error
...

PR #4152, opened 2019-10-15
The aria-label of the notification element should be localized strings.

huaiyinfeilong:
The aria-label of the notification element should be localized strings.
Blind people who are not native speakers of English will not be able to hear localized descriptions using a screen reader.


freeplant:
Thanks.

PR #4154, opened 2019-10-18
Enable single sign out when using OpenID Connect

MJochim:
This allows the administrator to add the setting OAUTH_END_SESSION_URL. If they do, when users log out, Seahub will try to log them out of their SSO session globally. This is a standard feature of OpenID Connect (so strictly speaking it’s not OAuth). See the [draft for session management](https://openid.net/specs/openid-connect-session-1_0-28.html) (technically it’s a draft and not yet a standard, but it is already deployed in OIDC software).

One problem is that the `http_or_https` variable is not set correctly in some cases (when deploying behind a reverse proxy that terminates the TLS connection). This is very likely handled correctly in other places in Seahub‘s code but I don’t know how.

### DIFF
--- a/seahub/auth/views.py
+++ b/seahub/auth/views.py
@@ -249,6 +249,19 @@ def logout(request, next_page=None,
             shib_logout_url += shib_logout_return
         response = HttpResponseRedirect(shib_logout_url)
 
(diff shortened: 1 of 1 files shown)


freeplant:
We will give it a check.

PR #4428, opened 2020-02-05
fix: group shares were not shown

silberzwiebel:
Code works on my deployment. It might have other not-thought-of-consequences, though. It's taken from here, so that everyone can profit from the bugfix: https://forum.seafile.com/t/error-in-seahub-when-accessing-group-shares/9540


freeplant:
This should have been fixed in the latest version.

PR #4442, opened 2020-02-14
Fix sharing with password.

### LINKED ISSUE #4440: Sharing with password and default path broken
This seems to be broken by https://github.com/haiwen/seahub/commit/44459bb36b8c3c3f3d57c9976357ea1cc4103439 [templates/share_access_validation.html](https://github.com/haiwen/seahub/blob/44459bb36b8c3c3f3d57c9976357ea1cc4103439/seahub/templates/share_access_validation.html#L15) stops checking whether `path` is empty and always append a `p=` parameter, which becomes `?p=` when the `path` is empty.

When the request is submitted.
https://github.com/haiwen/seahub/blob/82f16b9e47b41a73ed46311ebc6151745e3b6fc7/seahub/views/repo.py#L254-L255
`req_path` will be `''` (not handled by `get`) since it exists, and it caused a OOB error when trying to access the last character.

I now fixed this locally by changing the request handler into

```python
if req_path == '':
    req_path = '/'
elif req_path[-1] != '/':
    req_path += '/'
```

and it can probably also be fixed from the template as well. Should `?p=` be supported? Sumitting a PR for this should be easy but I don't know which one is the proper fix.

P.S. Is there a way to print out the backtrace when an error happens like this? I end up throwing an error in `server_error` (500 handler) and got the backtrace in `error.log`. Is there a better way?

Tested on 7.1.1.

yuyichao:
This implements both fixed mentioned in #4440 . Let me know if at least one of them isn't wanted for some reason...

Fix #4440

### DIFF
--- a/seahub/templates/share_access_validation.html
+++ b/seahub/templates/share_access_validation.html
@@ -4,7 +4,11 @@
 {% block main_panel %}
 <div class="mt-9 mb-4 mx-auto small-panel">
   <p class="intro">{% trans "Please input the password if you want to browse the shared file/directory." %}</p>
+  {% if path %}
   <form action="{% url view_name token %}?p={{ path }}" method="post" id="share-passwd-form">
+  {% else %}
+  <form action="{% url view_name token %}" method="post" id="share-passwd-form">
+  {% endif %}
     {% csrf_token %}
     <input type="hidden" name="token" value="{{ token }}" />
     <div class="form-group">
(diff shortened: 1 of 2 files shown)


yuyichao:
ping

yuyichao:
@llj @freeplant @LeoSirius @shuntian

freeplant:
Thanks for the PR.

We have looked into the problem and fixed the issue in https://github.com/haiwen/seahub/pull/4529/files

PR #4469, opened 2020-03-05
Fix video thumbnails generation

drizzt:
From 7.0 the video thumbnails generation is broken since the new frontend doesn't implement it.


Jonatino:
Will this get merged?

freeplant:
I'm afraid we will no longer support video thumbnail feature.

Jonatino:
@freeplant was there a specific reason for removing it? I really like that feature.

I had a bunch of problems with it until I removed the moviepy dependency and relied 100% on ffmpeg to do thumbnails. See #4688

PR #4483, opened 2020-03-17
fix no file history

jianhw:


jianhw:
[File history warning](https://trello.com/c/pBVUJ4yz/5627-file-history-warning)

Issue #4478, opened 2020-03-11
IndexError after trying to download files


1resu:
We want to share some files with a download link. After entering the password the following error message appears:

> Page unavailable
> Sorry, but the requested page is unavailable due to a server hiccup.
> Our engineers have been notified, so check back later.

This is the log output from seahub.log:

`2020-03-11 14:49:56,164 [ERROR] django.request:135 handle_uncaught_exception Internal Server Error: /seafile/d/8212a49539ab44f7adf8/
Traceback (most recent call last):
File "/seafile/seafile-server-7.1.2/seahub/thirdpart/django/core/handlers/exception.py", line 41, in inner response = get_response(request)
File "/seafile/seafile-server-7.1.2/seahub/thirdpart/django/core/handlers/base.py", line 249, in _legacy_get_response response = self._get_response(request)
File "/seafile/seafile-server-7.1.2/seahub/thirdpart/django/core/handlers/base.py", line 187, in _get_response response = self.process_exception_by_middleware(e, request)
File "/seafile/seafile-server-7.1.2/seahub/thirdpart/django/core/handlers/base.py", line 185, in _get_response response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/seafile/seafile-server-7.1.2/seahub/seahub/share/decorators.py", line 21, in _decorated    return func(request, fileshare, *args, **kwargs)
File "/seafile/seafile-server-7.1.2/seahub/seahub/share/decorators.py", line 67, in _decorated    return func(request, *args, **kwargs)
File "/seafile/seafile-server-7.1.2/seahub/seahub/views/repo.py", line 265, in view_shared_dir
if req_path[-1] != '/':
IndexError: string index out of range
`

thefkboss:
I have the same problem

t123yh:
Hi, this should be fixed by #4507. Could you please try it?

vinzBad:
The root error is in https://github.com/haiwen/seahub/blob/v7.1.3-server/seahub/views/repo.py#L256 it doesn't set the path for https://github.com/haiwen/seahub/blob/v7.1.3-server/seahub/templates/share_access_validation.html#L7

I fixed it in our instance by changing `{{ path }}` to `/` in `action="{% url view_name token %}?p={{ path }}"` in the file `seafile-server-7.1.3/seahub/seahub/templates/share_access_validation.html`

PR #4507, opened 2020-03-28
Use string.endswith to check path normalization.

t123yh:
When we encounter an empty path, like `https://demo.seafile.top/d/c56fc44545456ffea123/?p=`, the path string will have a length of 0. Using `path[-1]` index to check the last character will cause an internal server error (index out of bounds), so we probably should use `endswith` instead.

This should fix https://forum.seafile.com/t/internal-server-error-7-1-3/11368/16 and #4478.


freeplant:
The issue has been fixed in 8.0 release.

PR #4545, opened 2020-04-23
[shared video file view] add missing video-file-view class

blaniel:
The video-file-view css class is missing in the shared video file view so video-js player is not correctly resized.


freeplant:
Thanks. This is already fixed in the master.

PR #4586, opened 2020-06-12
filter file history list

jianhw:


jianhw:
[Expired history entries will not be deleted with Seaf-gc](https://trello.com/c/4H7zdaDY/5724-expired-history-entries-will-not-be-deleted-with-seaf-gc)

PR #4706, opened 2020-10-28
optimize delete expired invitations

jianhw:


jianhw:
[删除 pending invitation 的问题](https://trello.com/c/xfu4p50X/5845-%E5%88%A0%E9%99%A4-pending-invitation-%E7%9A%84%E9%97%AE%E9%A2%98)

PR #4717, opened 2020-11-13
Thumbnail support for HEIC/HEIF images

samuelduann:
HEIC/HEIF format is widely used among iOS users, but there's no thumbnail support for this format till now.  With Python library pyheif_pillow_opener, I found a simple and seamless way to integrate HEIF support to Pillow.

### DIFF
--- a/frontend/src/index.js
+++ b/frontend/src/index.js
@@ -10,8 +10,6 @@ import './assets/css/fontawesome.css';
 import './index.css';
 
 ReactDOM.render(
-  <I18nextProvider i18n={ i18n } >
-    <MarkdownEditor />
-  </I18nextProvider>,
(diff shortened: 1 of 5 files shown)


bigcat88:
Suggest you try `pillow-heif`. Supports encoding, decoding and `exif`, `xmp` for png/tiff/webp.
Registering it as a plugin for Pillow:

```python3
from pillow_heif import HeifImagePlugin
```

HadesD:
sorry, but why this still not being merged?

freeplant:
pyheif-pillow-opener is still in a very early beta version. We don't want to include it yet.

samuelduann:
> include

Thank you for the comment. Then what about pillow-heif? It looks being actively maintained.

freeplant:
[pillow-heif/](https://pypi.org/project/pillow-heif/) is in beta too and there are not many stars in the project.

PR #4775, opened 2020-12-29
Migrate PSD_tools to 1.8

nkelemen18:
Hi,

I've created a fix for PSD_tools package v1.8 or above according to PSD_tools migration manual.

PSDImage.load() is PSDImage.open() : [PSDImage](https://psd-tools.readthedocs.io/en/latest/migration.html#psdimage)
image.as_PIL() is image.compse() : [PIL Export](https://psd-tools.readthedocs.io/en/latest/migration.html#pil-export)


olof-nord:
I am not the repo owner or a maintainer, but in general, the title of thie PR seems to indicate that a version of a dependency has increased. The diff does not show any package/dependency change, but instead adjusts two function calls - perhaps you could clarify a bit more what exactly is changed here: and perhaps more importantly also why? @nkelemen18

PR #4816, opened 2021-02-10
open collabora view, fixes #4627

### LINKED ISSUE #4627: Seahub shared link previews for Collabora Online do not work
When creating a share link for document files, you can select the "Preview only" or "Preview and download" permissions. Unfortunately, this preview does not actually work if Collabora Online integration is used. Instead, it gives this error message: https://github.com/haiwen/seahub/blob/master/seahub/views/file.py#L1229

The reason for this is the `get_wopi_dict` method returning `None` (https://github.com/haiwen/seahub/blob/master/seahub/wopi/utils.py#L131). This method returns `None` because no action named `view` matching the file type is found in the WOPI `discovery.xml`. 

The Collabora Online `discovery.xml` can be found here: https://github.com/LibreOffice/online/blob/master/discovery.xml. As you can see, for example for an ods file (our main use-case), only the `edit` action is specified: https://github.com/LibreOffice/online/blob/master/discovery.xml#L29. However, Collabora Online does in fact also support the `view` action (read-only). I assume the developers created the implicit relationship that, if a file type supports the `edit` action, it also supports the `view` action. Consequently, they did not explicitly specify the `view` action for a lot of file types.

As a solution, we created our own `discovery.xml` in `/etc/loolwsd/` and duplicated every `edit` action for a file type with a `view` action. Unfortunately, this is far from perfect. It might be better if seahub also implicitly supports previews for any file if `edit` is supported (with Collabora Online integration). Of course, care should be taken to test every file type and ensure that read-only mode actually works.

We run Seafile Server Version: 7.1.4.

vali-um:
This fixes #4627
collabora online provides only one action to edit an file extension and implies that view is also possible.
so if the action allows edit we also use the provided action to view


vali-um:
sorry i mixed up some things here.
closing this an will open a new and clean PR

PR #4832, opened 2021-03-12
allow superuser creation without password prompt

Wurzelmann:
This patch in combination with another little patch in seafile-server allows the creation of a superuser without a password prompt.


freeplant:
I don't understand why this feature is needed.

Wurzelmann:
We needed this functionality in order to deploy a large number of seafile instances at once.

PR #4849, opened 2021-03-23
admin rename department

imwhatiam:


seahub/api2/endpoints/admin/groups.py
```diff
@@ -179,6 +181,7 @@ def put(self, request, group_id):
 
         1. transfer a group.
         2. set group quota
+        3. rename group quota
```
freeplant, line 184: rename group, 去掉 quota

PR #4852, opened 2021-03-29
fix onlyoffice cache key

blaniel:
Cache key is based on repo_id and file_path. When sharing a folder, repo_id is
not the same as the owner's resulting in diverging onlyoffice sessions and thus,
loss of data.
Use origin_repo_id and origin_repo_file_path to have consistent keys.


freeplant:
Thanks. It has been added in this PR: https://github.com/haiwen/seahub/pull/4925

PR #4858, opened 2021-03-31
fix onlyoffice cache timeout

blaniel:
When opening a document, the cache timeout is set to None so the key never
expires. However when force save is received, the cache timeout is not set so
it defaults to 300. When a new user opens the document after this timeout, a
new cache key is generated and a new editing session is created.

Set cache timeout to None in onlyoffice_editor_callback so cache key never
expires.


freeplant:
Thank you very much. The fix will be added in the next release.

freeplant:
Thanks. It has been added in this PR: https://github.com/haiwen/seahub/pull/4925

PR #4905, opened 2021-05-25
show filename as title when view published file

imwhatiam:


seahub/templates/wiki/wiki.html
```diff
@@ -4,6 +4,8 @@
 {% render_bundle 'wiki' 'css' %}
 {% endblock %}
 
+{% block sub_title %} {{filename}} - {% endblock %}
```
freeplant, line 7: 这里把 filename 中的文件名后缀去掉吧

PR #4930, opened 2021-07-08
Correction spelling swedish mindra to mindre

thuma:
Error in the Swedish translation.


olof-nord:
As someone who also know Swedish, I can confirm this change is correct.

freeplant:
The translation need to be corrected in Seafile's translation project at Transifex platform (https://www.transifex.com/projects/p/seahub/).

PR #4995, opened 2021-10-01
Add SITE_DESCRIPTION config variable

moschlar:
Which gets used in the html header `<meta name="description" ... />` tag.


freeplant:
Merged. Thanks.

PR #4996, opened 2021-10-01
Use comma as html meta keyword separator

moschlar:
As per
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name this
it the correct character to separate the keywords listed there.

I also updated some of the translation strings (but not those with
non-latin character sets, sorry).


imwhatiam:
@moschlar Hello.

We use Transifex to make Seafile internationalization, so I closed this pr and opened a new one, https://github.com/haiwen/seahub/pull/5044,.

As for the translation strings we will update them on Transifex.

PR #4998, opened 2021-10-09
CUSTOM_FAVICON_PATH should be a path to a directory

embeddedc:
CUSTOM_FAVICON_PATH should be a path to a directory to allow modern favicon handling


imwhatiam:
@embeddedc Thanks for this pr.

If we change CUSTOM_FAVICON_PATH to a directory, it will destroy those users who have already configured it.

So we will close this pr and open a new one https://github.com/haiwen/seahub/pull/5049/files

PR #5025, opened 2021-11-09
Transfer single repo shares when migrating user

moschlar:


seahub/api2/endpoints/admin/account.py
```diff
@@ -92,6 +92,11 @@ def post(self, request, email, format=None):
             for r in seafile_api.get_owned_repo_list(from_user):
                 seafile_api.set_repo_owner(r.id, user2.username)
 
+            # transfer shared repos to new user
+            for r in seafile_api.get_share_in_repo_list(from_user, 0, -1):
```
moschlar, line 96: Or should the second parameter be `-1 `, too?

imwhatiam: Yes, it should be -1

Issue #5062, opened 2021-12-15
Collaboration notices get resent on every interval if user has contact_email != username


moschlar:
Because the initial query is done with the username and the timestamp of the sent notices is done with the contact_email.
I'll whip up a quick PR.

moschlar:
Would you mind including this bugfix (also for version 8), please?

PR #5063, opened 2021-12-15
Fix collaboration notice sending (Closes: #5062)

moschlar:
By setting `to_user = contact_email`, the timestamp `collaborate_last_emailed_time` gets set for the contact email instead of the username, but that is getting used initially for querying the last emailed timestamps. This leads to notification emails getting sent every time the email sending job runs.

This change makes it more in sync with send_file_updates.py, too.


moschlar:
Ping!

freeplant:
Sorry for missing the PR. We will give it a check.

PR #5066, opened 2021-12-17
exec 'portal:logout' AscDesktopEditor command

imwhatiam:


LinneyS (approved):
We have tested this in ONLYOFFICE Desktop Editors. Works great
Thank you

PR #5150, opened 2022-03-06
Use full URI in oauth authorization_response

### LINKED ISSUE #5083: OAuth 2 MUST utilize https
According to this topic https://forum.seafile.com/t/oauth-fails-with-must-utilize-https-but-everything-is-https/10401

OAuth fails with

> [ERROR] seahub.oauth.views:150 oauth_callback (insecure_transport) OAuth 2 MUST utilize https.

https://github.com/haiwen/seahub/blob/e92fcd44e3e46260597d8faa9347cb8222b8b10d/seahub/oauth/views.py#L133
request.get_full_path() just containes /seafile/oauth/… and not the full URL.

masonh:
authorization_response parameter should be "The full URL of the redirect back to you" because it checks for secure transport.

fixes #5083

### DIFF
--- a/seahub/oauth/views.py
+++ b/seahub/oauth/views.py
@@ -130,7 +130,7 @@ def oauth_callback(request):
         token = session.fetch_token(
             TOKEN_URL,
             client_secret=CLIENT_SECRET,
-            authorization_response=request.get_full_path())
+            authorization_response=request.build_absolute_uri())
 
         if 'user_id' in session._client.__dict__['token']:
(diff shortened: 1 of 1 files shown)


masonh:
tested on my home instance (9.0.4 CE docker)

masonh:
addressed by PR https://github.com/haiwen/seahub/pull/5231 / commit fc04a91eea995391c0a45b6f1089950512f289c6

PR #5195, opened 2022-06-07
Fix share links with OnlyOffice 7.1

dani:
mode must be either 'edit' or 'view'. Prevents auth missing required
parameter editorConfig.mode (since 7.1 version) error when opening a
share link pointing on an office document

See https://forum.seafile.com/t/seafile-9-onlyoffice-7-1-cant-open-publicly-shared-docs/16521


flowjzh:
I can confirm that the issue described in the forum link exists with CE v9.0.4. This change got it fixed.

PR #5217, opened 2022-07-21
Fixed preview PSD images

snk26:
Fixed PSD image preview as methods in psd_tools has been updated.
https://psd-tools.readthedocs.io/en/latest/migration.html


snk26:
The problem can be seen in the demo seafile
https://demo.seafile.com/lib/40a6a384-3358-4cba-8faa-9ea59346a35d/file/02_Home.psd

freeplant:
Thanks!

PR #5274, opened 2022-10-05
Fix sqlite2mysql script problems with seahub_db

n00b42:
This MR fixes problems of the sqlite2mysql.py toolwith "IF NOT EXISTS" statements and the value of "DEFAULT" statements.

Fixes https://github.com/haiwen/seafile/issues/2505

### DIFF
--- a/scripts/sqlite2mysql.py
+++ b/scripts/sqlite2mysql.py
@@ -41,7 +41,7 @@ for line in fileinput.input():
     if re.match(r'^CREATE TABLE.*', line):
         searching_for_end = True
 
-    m = re.search('CREATE TABLE [`"]?(\w*)[`"]?(.*)', line)
+    m = re.search('CREATE TABLE(?: IF NOT EXISTS)? [`"]?(\w+)[`"]?(\s*\(.*)', line)
     if m:
         name, sub = m.groups()
         sub = sub.replace('"','`')
--- a/scripts/sqlite2mysql.py
+++ b/scripts/sqlite2mysql.py
@@ -67,6 +67,9 @@ for line in fileinput.input():
         # replace " and ' with ` because mysql doesn't like quotes in CREATE commands
         line = line.replace('"', '`').replace("'", '`')
 
+    # fix default values parantheses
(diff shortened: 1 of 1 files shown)


imwhatiam:
@n00b42

imwhatiam:
This two table's sqlite3 creation sentence is:

```
CREATE TABLE IF NOT EXISTS "api2_tokenv2" ("key" varchar(40) NOT NULL PRIMARY KEY, "user" varchar(255) NOT NULL, "platform" varchar(32) NOT NULL, "device_id" varchar(40) NOT NULL, "device_name" varchar(40) NOT NULL, "platform_version" varchar(16) NOT NULL, "client_version" varchar(16) NOT NULL, "last_accessed" datetime NOT NULL, "last_login_ip" char(39) NULL, "created_at" datetime NOT NULL, "wiped_at" datetime NULL);
```

and 

```
CREATE TABLE IF NOT EXISTS "share_fileshare" ("id" integer NOT NULL PRIMARY KEY AUTOINCREMENT, "username" varchar(255) NOT NULL, "repo_id" varchar(36) NOT NULL, "path" text NOT NULL, "token" varchar(100) NOT NULL UNIQUE, "ctime" datetime NOT NULL, "view_cnt" integer NOT NULL, "s_type" varchar(2) NOT NULL, "password" varchar(128) NULL, "expire_date" datetime NULL, "permission" varchar(50) NOT NULL);
```

no `default` value should be replaced?

So I don't get `fix default value quotation` commit's usage, can you explain more?

n00b42:
It was a quite old install, and according to https://github.com/haiwen/seahub/blob/6ddaf4e47da2b4acb7b7abe56972e3809de0b195/scripts/upgrade/sql/6.0.0/sqlite3/seahub.sql#L24 the "default" value was/is added to the table in sqlite....

imwhatiam:
Thanks for the additional info.

But when I tested

```
line = re.sub(r"default `([^`]*)`", "default '\1'", line, 0, re.IGNORECASE)
```
image


the string
```
created_at datetime NOT NULL DEFAULT `1970-01-01 00:00:00`,
```

was changed to
```
created_at datetime NOT NULL default '\x01',
```
image

Is `\1` a misspell?

PR #5280, opened 2022-10-10
added missing blanks

itsNik:
Added missing blanks in German translation.


freaky-m0:
One year and no merge for such a simple translation fix? :D

freeplant:
Hi all, the translation need to be done in Transifex platform. Merging in code directly will be overwritten by the translation downloaded from Transifex.

PR #5324, opened 2022-12-02
replace non rfc 6761 domain

mainboarder:


mainboarder:
integrated into https://github.com/haiwen/seahub/pull/5320

PR #5549, opened 2023-07-15
Update avatar/settings.py to update deprecated PIL statement

lkuza2:
ANTIALIAS was deprecated in PIL and has been removed in the latest version (which will cause errors for Seahub to start)

Image.LANCZOS provides the same algorithm and is the recommended statement


edacval:
Same change needs to be applied to thumbnail/utils.py too, otherwise image prewiev will not work.  Just not sure about the right choice `Image.LANCZOS` vs `Image.Resampling.LANCZOS`

freeplant:
The problem has been fixed in https://github.com/haiwen/seahub/pull/5570

PR #5551, opened 2023-07-17
Update settings.py to reflect removed ANTIALIAS in PIL

iskur:
See https://github.com/mpetroff/pannellum/issues/1119


iskur:
I see this has a duplicate, sorry!
https://github.com/haiwen/seahub/pull/5549

PR #5647, opened 2023-09-15
Fix spelling error in czech locales

kepi:
Fix spelling error


freeplant:
Please fix it in the Transifex platform. Otherwise, it will be overwritten when the next time translation are synced from Transifex.

kepi:
ok, i sent request to join

PR #5948, opened 2024-02-07
update German translation

likesclever:


freeplant:
Fix issue: https://github.com/haiwen/seahub/issues/5950

PR #6131, opened 2024-05-29
support comma-separated value for seafile_groups claim

thewilli:
IDPs like Entra ID don't support multi-value claims except for group membership, but those values don't match up easily with Seafile group IDs.

This commit support a single-value input for seafile_groups to contain commas as separator, so it can easily be used with Entra ID static values and transformations like `Join()`.


freeplant:
Thank you

PR #6158, opened 2024-06-01
Append extra settings for seahub

xxbillouxx:
If `seahub_settings_extra.py` exists, it is automatically added at the end of the generated `seahub_settings.py`

By this way, it is no more required : 
- to write manually the extra settings at the end of the `seahub_settings.py` 
- and to restart the service

as described [here](https://manual.seafile.com/deploy/only_office/#configure-seafile-server).

This is useful when seafile is deployed with a docker compose file: instead of starting the container and manually doing a `docker exec` to edit the `seahub_settings.py` inside the container and restarting the service, we just provide a seahub_settings_extra.py with the docker 'volume' feature.

### DIFF
--- a/scripts/setup-seafile-mysql.py
+++ b/scripts/setup-seafile-mysql.py
@@ -1030,6 +1030,7 @@ class SeahubConfigurator(AbstractConfigurator):
         self.admin_email = ''
         self.admin_password = ''
         self.seahub_settings_py = os.path.join(env_mgr.central_config_dir, 'seahub_settings.py')
+        self.seahub_settings_extra_py = os.path.join(env_mgr.central_config_dir, 'seahub_settings_extra.py')
 
     def ask_questions(self):
         pass
@@ -1045,6 +1046,7 @@ class SeahubConfigurator(AbstractConfigurator):
             fp.write('SERVICE_URL = "http://%s"' % ccnet_config.ip_or_domain)
             fp.write('\n')
             self.write_database_config(fp)
+            self.write_extra_config(fp)
 
     def write_utf8_comment(self, fp):
         fp.write('# -*- coding: utf-8 -*-')
@@ -1080,6 +1082,14 @@ class SeahubConfigurator(AbstractConfigurator):
                                port=db_config.mysql_port)
 
         fp.write(text)
+
+    def write_extra_config(self, fp):
+        try:
+            with open(self.seahub_settings_extra_py) as fp_extra:
+                fp.write('\n')
+                fp.writelines(fp_extra.readlines())
+        except FileNotFoundError:
+            pass
 
     def ask_admin_email(self):
         print()


freeplant:
seahub_extra_settings.py are no longer used anymore since version 10.0.

In which case do you need this file?

xxbillouxx:
Hi @freeplant 

Thanks for your feedback. 

My `seahub_settings_extra.py` is a new file that I created, there is no link with the `seahub_extra_settings.py` that existed in the previous versions of Seafile.
I realize my new file has a name that could be confusing with the old `seahub_extra_settings.py` 😕 

I re-explain below the context for which I created this file:

I use the `docker-compose.yml` (as detailed [here](https://manual.seafile.com/docker/deploy_seafile_with_docker/#download-and-modify-docker-composeyml)) to create a Seafile container and all its related containers (database and memcached).

And I have another `docker-compose.yml` (on another server) to create an Onlyoffice container.

As specified [here](https://manual.seafile.com/deploy/only_office/#configure-seafile-server) , to be able to link a Seafile server to a Onlyoffice server:
- I must add some extra configuration in the `seahub_settings.py` 
- and I need to restart the Seafile server after.

=> this add some manual operations. 

I would like to avoid these manual steps and do all of that automatically at the Seafile container startup.

So this is why I would like to do that:
- creating a file `seahub_settings_extra.py`
```
# Enable OnlyOffice
ENABLE_ONLYOFFICE = True
VERIFY_ONLYOFFICE_CERTIFICATE = False
ONLYOFFICE_APIJS_URL = 'http{s}://{your OnlyOffice server's domain or IP}/web-apps/apps/api/documents/api.js'
ONLYOFFICE_FILE_EXTENSION = ('doc', 'docx', 'ppt', 'pptx', 'xls', 'xlsx', 'odt', 'fodt', 'odp', 'fodp', 'ods', 'fods')
ONLYOFFICE_EDIT_FILE_EXTENSION = ('docx', 'pptx', 'xlsx')

# Enable force save to let user can save file when he/she press the save button on OnlyOffice file edit page.
ONLYOFFICE_FORCE_SAVE = True

# if JWT enabled
ONLYOFFICE_JWT_SECRET = 'your-secret-string'
```

- adding in the docker compose:
```
  seafile:
    image: seafileltd/seafile-mc:11.0-latest
    [...]
    volumes:
      [...] 
      - ./seahub_settings_extra.py:/shared/seafile/conf/seahub_settings_extra.py
    [...]
```

So at the Seafile container startup, the `setup-seafile-mysql.py` will check if there is an `seahub_settings_extra.py` and will append its content to the end of the `seahub_settings.py`.

This new `seahub_settings_extra.py` could be used for any customization that requires today to manual edit the `seahub_settings.py` 😃 : no more manual editing, no need to restart the Seafile server anymore.

Let me know if it makes sense.

Thank you.

freeplant:
Thanks for the explanation. I see your logic now. We will think about it.

PR #6268, opened 2024-06-28
fix system-admin user page bug

Michael18811380328:


frontend/src/components/single-selector.js
```diff
@@ -67,7 +67,7 @@ class Selector extends Component {
     return (
       <div className="sf-single-selector position-relative">
         <span className="cur-option" onClick={this.onToggleClick}>
-          {currentSelectedOption.text}
+          {currentSelectedOption ? currentSelectedOption.text : ''}
```
Michael18811380328, line 70: when currentSelectedOption is null and there is a bug

PR #6608, opened 2024-08-21
fix: content style

YangGuoXuan-0503:


Michael18811380328:
This PR fix two issues:
1. file page double border
2. Wiki page dropdown menu

PR #6615, opened 2024-08-22
force kill wsgidav

imwhatiam:


laur89:
Why do we even need to kill wsgidav from `seafile.sh`? The script already sends sigterm to controller:
```sh
pkill -SIGTERM -f "seafile-controller -c ${default_ccnet_conf_dir}"
```
 which in turn causes [controller to send sigkill to webdav](https://github.com/haiwen/seafile-server/blob/1afed5d770a8dbfb6a4b383bf33ca233e16bec30/controller/seafile-controller.c#L530).

If anything, we should fix how controller terminates webdav.

freeplant:
Sometimes the wsgidav cannot be shutdown correctly. Anyway, we will move the control part of  wsgidav to monitor.sh instead of seafile-controller in version 12. We will recheck the issue later.

PR #6956, opened 2024-10-28
Add possibility to import LDAP account from RemoteUserBackend

Josue-T:
# Problem

Sometime we use the Remote user backend authentication to authenticate automatically the user. But we also use LDAP to store all user info and also in case of the user don't use the SSO to authenticate (by example from the mobile or desktop app).

This provide the possibility to import the LDAP account (as optional), if a remote user header is provided and an account don't already exist for this user.

Note I'll need this patch for the [Yunohost seafile package](https://github.com/YunoHost-Apps/seafile_ynh/), to have a clean fix of https://github.com/YunoHost-Apps/seafile_ynh/issues/128.

### DIFF
--- a/seahub/auth/backends.py
+++ b/seahub/auth/backends.py
@@ -9,7 +9,7 @@ from django.conf import settings
 
 from seaserv import ccnet_api, seafile_api
 
-from seahub.base.accounts import User, AuthBackend
+from seahub.base.accounts import User, AuthBackend, CustomLDAPBackend
 from seahub.profile.models import Profile
 from seahub.utils.file_size import get_quota_from_string
 from seahub.role_permissions.utils import get_enabled_role_permissions_by_role
@@ -93,9 +93,14 @@ class SeafileRemoteUserBackend(AuthBackend):
     create_unknown_user = getattr(settings, 'REMOTE_USER_CREATE_UNKNOWN_USER',
                                   True)
 
+    import_from_ldap = getattr(settings, 'REMOTE_USER_IMPORT_FROM_LDAP',
+                               True)
+
     # Create active user by default.
     auto_activate = getattr(settings,
                             'REMOTE_USER_ACTIVATE_USER_AFTER_CREATION', True)
+
+    ldapBackend = CustomLDAPBackend()
 
     # map user attribute in HTTP header and Seahub user attribute
     # REMOTE_USER_ATTRIBUTE_MAP = {
@@ -143,6 +148,9 @@ class SeafileRemoteUserBackend(AuthBackend):
 
         # get user from ccnet
         user = self.get_user(username)
+        if not user and self.import_from_ldap:
+            user = self.ldapBackend.authenticate(ldap_user=username, password=None, no_check_password=True)
+
         if not user:
             # when user doesn't exist
             if not self.create_unknown_user:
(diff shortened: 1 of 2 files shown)


imwhatiam:
@Josue-T 
Hello, in Seafile version 11, we changed the user identity, which you can see here: [Change of User Identity](https://manual.seafile.com/latest/upgrade/upgrade_notes_for_11.0.x/#change-of-user-identity). For users integrated with LDAP, please also refer to this guide: [Reimplementation of LDAP Integration](https://manual.seafile.com/latest/upgrade/upgrade_notes_for_11.0.x/#reimplementation-of-ldap-integration) for further actions.

If these instructions don’t resolve your issue, we can look further into how to solve it.

Josue-T:
Hello,

I had this issue since we upgraded to Seafile 11 and we updated the LDAP configuration accordingly. So it's more since we upgraded to Seafile 11 and applied the upgrade instruction that we have this issue.

Note you can check the used seahub config file here: https://github.com/YunoHost-Apps/seafile_ynh/blob/master/conf/seahub_settings.py

imwhatiam:
Hello, as mentioned here,
image

starting from version 11, there should now be two types of users in Seafile:

1. **Old Users**: These have user IDs in email format (e.g., `old-user@example.com`) and are stored in the `ccnet_db.EmailUsers` table.

2. **New Users**: Newly created users have user IDs in virtual ID format (e.g., `adc023e7232240fcbb83b273e1d73d36@auth.local`). This virtual ID is not exposed to the user. It’s also stored in `ccnet_db.EmailUsers`, while a mapping between this virtual ID and the user’s contact email (e.g., `new-user-contact-email@example.com`) is stored in the `seahub_db.profile_profile` table.

All Seafile users are stored in the `ccnet_db.EmailUsers` table. User login relies on finding the user in this table:

1. If the user is an **old user** and directly exists in `ccnet_db.EmailUsers`, they can log in immediately.

2. If the user is a **new user**, they log in by entering their contact email (e.g., `new-user-contact-email@example.com`). Since this email isn’t in `ccnet_db.EmailUsers`, Seafile needs to look it up in `seahub_db.profile_profile`, retrieve the virtual ID (`adc023e7232240fcbb83b273e1d73d36@auth.local`), and then log in the user.

The issue you’re encountering is likely due to a missing lookup in `seahub_db.profile_profile` within `SeafileRemoteUserBackend`.

Based on the analysis above, I wrote some debug code. You can deploy it on your server and check the output in `seahub.log`.

https://github.com/haiwen/seahub/commit/1798772eeccc510cfdef6598c49ab8c0070501aa

Josue-T:
Hello,

Thanks for all of theses explanation. Just to be sure that we understood correctly the issue,  we have 2 way to authenticate on seafile in the Yunohost package.

The fist way is by LDAP. This can be the case if the user is not logged in the SSO or if the user authenticate from a application (like the desktop client or the mobile app).

And if the user is authenticated in the Yunohost SSO, the user is authenticated by the RemoteUserBacked. But if the user do the first login by the SSO, I wan't the user account is imported from LDAP and not just created from the RemoteUserBacked which provide less account information (like user full name). So I put  `REMOTE_USER_CREATE_UNKNOWN_USER = False` in the seahub configuration as to me it's not a unknown user it's the user from LDAP.

So in the current implementation, is it already expected that a not imported user from LDAP will be imported if the user is logged in by the RemoteUserBacked even if the `REMOTE_USER_CREATE_UNKNOWN_USER = False` is set ?

If it's the case, I'll do the debug as you suggested.

imwhatiam:
If `REMOTE_USER_CREATE_UNKNOWN_USER = False`, SeafileRemoteUserBackend will not create a new user; instead, it will display the message "Error, new user registration is not allowed, please contact administrator." Therefore, when a user logs in through SeafileRemoteUserBackend with `REMOTE_USER_CREATE_UNKNOWN_USER = False`, the user must already exist in Seafile.

--

If you want to obtain more user information when users log in via Yunohost SSO -> SeafileRemoteUserBackend, you can achieve this by configuring `REMOTE_USER_ATTRIBUTE_MAP`. This configuration maps certain headers in the HTTP request to user attributes in Seafile, allowing for the updating of user attributes.

For example:

```
REMOTE_USER_ATTRIBUTE_MAP = {
    'HTTP_DISPLAYNAME': 'name',
    'HTTP_MAIL': 'contact_email',
}
```

In this configuration, `HTTP_DISPLAYNAME` is an HTTP header that maps to the `name` attribute of a Seafile user, while `HTTP_MAIL` is an HTTP header that maps to the `contact_email` attribute of a Seafile user.

Josue-T:
Thank you for the explantation. So closing it as your approach around this is a bit different than what I thought.

PR #7011, opened 2024-11-08
fix recent used item bug

Michael18811380328:


Michael18811380328:
This issue is fixed in PR https://github.com/haiwen/seahub/pull/7011

PR #7109, opened 2024-11-26
Lib settings fixup

llj:


Michael18811380328:
This PR fixed the issue of inconsistent IDs in the library settings dialog(face panel and tag panel id inconsistent), adjusted some code structure, removed redundant code, and tested normally after modification.

PR #7135, opened 2024-12-02
fix delete all departements

Michael18811380328:


frontend/src/pages/sys-admin/departments-v2/departments-v2.js
```diff
@@ -190,11 +190,18 @@ class DepartmentsV2 extends React.Component {
         let rootNodes = this.state.rootNodes.slice(0);
         let rootIndex = rootNodes.findIndex(node => node.id === operateNode.id);
         rootNodes.splice(rootIndex, 1);
-        this.setState({
-          rootNodes,
-          checkedDepartmentId: rootNodes[0].id,
```
Michael18811380328, line 195: Seahub support no rootNodes, so when delete last department, rootNodes[0] is null, id can not found.

PR #7139, opened 2024-12-03
fix get image index bug

Michael18811380328:


frontend/src/components/dirent-list-view/dirent-list-view.js
```diff
@@ -233,7 +233,7 @@ class DirentListView extends React.Component {
     this.setState({
       isImagePopupOpen: true,
       imageItems: imageItems,
-      imageIndex: items.indexOf(curItem)
+      imageIndex: items.findIndex((item) => item.id === curItem.id),
```
Michael18811380328, line 236: note:
Arrsy.indexOf() is suitable for simple value searches using strict equality.
Arrsy.findIndex() is suitable for complex searches using a callback function.
If you need to perform a simple value search, indexOf is a better choice.
If you need to perform a complex search, findIndex is a better choice.
In this area, we should compare clicked dirent. It is a complex object and indexOf() can get error value, so use findIndex() instead.

PR #7327, opened 2025-01-07
Share to user support select department

imwhatiam:


frontend/src/components/dialog/department-detail-dialog.js
```diff
@@ -0,0 +1,304 @@
...
+  resetCurrentPage = () => {
+    this.setState({ currentMemberPage: 1, hasMore: true });
+  };
+
+  getOrgMembers = async () => {
+    const orgID = window.app.pageOptions.orgID;
+    await this.resetCurrentPage();
+    this.setState({ membersLoading: true });
+    let currentMemberPage = this.state.currentMemberPage;
+    seafileAPI.getOrganizationMembers(orgID, currentMemberPage).then((res) => {
```
Michael18811380328, line 141: 还需要加一个 getOrganizationMembers 获取 org 成员的 API

frontend/src/components/dialog/department-detail-dialog.js
```diff
@@ -0,0 +1,304 @@
...
+      if (!error.response || error.response.status !== 403) {
+        toaster.danger(errMsg);
+      }
+    });
+  };
+
+  getMoreOrgMembers = async () => {
+    const orgID = window.app.pageOptions.orgID;
+    let currentMemberPage = this.state.currentMemberPage;
+    seafileAPI.getOrganizationMembers(orgID, currentMemberPage).then((res) => {
```
Michael18811380328, line 162: 同上

Michael18811380328:
- Shen share the library to other users, user can be selected from the department
- When add group members, users can be selected from the department
Note: When selecting department users, both regular users or org users, only the departments and their members that have been set up by the system administrator are listed.

PR #7460, opened 2025-02-14
clean image preview cache after rotate

Michael18811380328:
After rotating the image, avoid using browser cache and load a new image from the server:
1. First loading: Before opening an image, we don't know if it has been rotated or if there is an old image cache locally, so we initialize it with a timestamp to ensure that the existing cache is cleared.
2. After rotating the image: reset the timestamp to ensure that the latest thumbnail is obtained.


Michael18811380328:
Use another method to clean cache https://github.com/haiwen/seahub/pull/7554 , so close this PR

PR #7472, opened 2025-02-17
[13.0] fix dropdown menu style

Michael18811380328:


frontend/src/components/cur-dir-path/dir-path.js
```diff
@@ -2,7 +2,7 @@ import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { UncontrolledTooltip } from 'reactstrap';
 import { Link } from '@gatsbyjs/reach-router';
-import DirOperationToolBar from '../../components/toolbar/dir-operation-toolbar';
+import DirOperationToolbar from '../../components/toolbar/dir-operation-toolbar';
```
Michael18811380328, line 5: change export Name is same with import Name

frontend/src/components/dropdown-menu/item-dropdown-menu.js
```diff
@@ -224,7 +223,7 @@ class ItemDropdownMenu extends React.Component {
                       <i className="sf3-font-down sf3-font rotate-270"></i>
                     </DropdownToggle>
                     <DropdownMenu
-                      positionFixed={true}
+                      className="position-fixed"
```
Michael18811380328, line 226: reactstrap 9.x Dropdown not support positionFixed props, so use className instead

PR #7475, opened 2025-02-17
add image deletion button on table view image previewer

Aries-0331:


frontend/src/metadata/views/table/table-main/records/body.js
```diff
@@ -563,6 +563,7 @@ class RecordsBody extends Component {
             modifyColumnData={this.props.modifyColumnData}
             getTableCanvasContainerRect={this.props.getTableCanvasContainerRect}
             updateFileTags={this.props.updateFileTags}
+            deleteRecords={this.props.deleteRecords}
```
Michael18811380328, line 566: 元数据视图，应该判断资料库的权限

PR #7851, opened 2025-05-23
fix permission

Aries-0331:


Michael18811380328:
Fix readonly repo can not modify in metadata views.

PR #8935, opened 2026-05-05
Add REDIS_DB env var for cache db selection

buttilda:
Allows seahub's redis cache to connect to a non-zero Redis DB index.
Use case: shared Redis instance with other things.

Default behavior is unchanged. When `REDIS_DB` is unset (or empty), the constructed URL is identical to the one before this change, and redis-py falls back to db 0.

Tested locally: REDIS_DB=2 produces LOCATION = 'redis://redis:6379/2'.


buttilda:
in case you haven't seen it... pinging @shuntian and @Aries-0331 as you two seem active here

freeplant:
I'm sorry we don't have a plan to support such use case at this moment. You can post a feature request in our forum.

buttilda:
alright.
thanks for the reply

PR #9051, opened 2026-05-27
fix: invisible-folder filter must use a joined path and exact-prefix …

sdsys-ch:
…match

The metadata endpoints construct the path passed to is_invisible_path by string-concatenating `_parent_dir + _name`. With `_parent_dir` stored as `/Folder` (no trailing slash, observed in production), this produces `/Folderfile.txt` — no `/` separator — and the prefix match in is_invisible_path then either fails (when invisible_path is stored with a trailing slash, so the filter never matches and files inside leak) or over-matches (when stored without, so a sibling like `/Folderbackup/` also gets hidden).

Two changes:

- MetadataRecords.get and MetadataRecord.get now use posixpath.join to build the path being checked, which produces a proper file path regardless of trailing-slash convention in storage.
- is_invisible_path normalizes the stored invisible path (strips trailing slash) and requires an exact match or a `/`-bounded prefix match. This is tolerant of either trailing-slash convention and eliminates the sibling-prefix over-match.

MetadataRecord.get also filters query_result['results'] in-place rather than into a local that was previously not used in the response, matching the sibling MetadataRecords.get pattern.

Adds unit tests covering: file inside invisible folder (both storage conventions), nested file, sibling-prefix folder (regression for the over-match bug), no-separator concat (regression for the under-match bug), exact directory match, multiple invisible folders, unknown repo, and empty invisible set.


sdsys-ch:
@r350178982 : This is a follow-up to the filter added in commit [773d4f17](https://github.com/haiwen/seahub/commit/773d4f17). Thank you for acting on this, highly appreciated!

That commit wired up `is_invisible_path` on `MetadataRecords.get` and `MetadataRecord.get`,  but the path argument was built by string-concatenation (`_parent_dir + _name`). With production storage holding `_parent_dir` without a trailing slash, the resulting `/Folderfile.txt` no longer matches the invisible-path prefix, so files inside invisible folders are still returned to users with `invisible` permission. Reproduced on our local instance running latest v13 in docker.

The PR replaces the concat with `posixpath.join` and hardens `is_invisible_path` against prefix-overmatching (`/HRbackup/file.txt` previously matched invisible `/HR`).

Disclaimer: POC and analysis was guided by claude opus 4.7 for this PR. Code tested and refined locally by a human. POC available on request.

PR #9052, opened 2026-05-27
fix: revert permission check should target file/dir path, not library root

sdsys-ch:
FileRevert.put and DirRevert.put hard-code the permission check to the library root (`check_folder_permission(request, repo_id, '/')`). A user shared into a sub-folder with `rw` (no library-level `rw`-access) is incorrectly denied revert on files and directories they otherwise have full write access to, while every sibling write operation in FileView and DirView checks the file/dir's own location.

- FileRevert.put now checks the file's parent_dir, matching FileView.delete.
- DirRevert.put now checks the directory path itself, matching DirView.delete.

check_folder_permission resolves inherited share permissions downward, so a user with `rw` on `/Shared/ProjectA/` is correctly granted access for any descendant, while a user with no library or relevant folder share still gets denied — the only behavior change is that folder-rw shares grant revert as they grant every other write operation.

### DIFF
--- a/seahub/api2/views.py
+++ b/seahub/api2/views.py
@@ -3563,7 +3563,8 @@ class FileRevert(APIView):
             error_msg = 'file %s not found.' % path
             return api_error(status.HTTP_404_NOT_FOUND, error_msg)
 
-        if check_folder_permission(request, repo_id, '/') != 'rw':
+        parent_dir = os.path.dirname(path)
+        if check_folder_permission(request, repo_id, parent_dir) != 'rw':
             error_msg = 'Permission denied.'
             return api_error(status.HTTP_403_FORBIDDEN, error_msg)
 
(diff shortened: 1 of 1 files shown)


sdsys-ch:
Noticed this in local testing. Every other write operation on `seahub/api2/views.py:FileView` and `DirView` correctly checks the file's or directory's own location.

Reproduced on Seafile Pro 13.0.x with two SAML-provisioned users. User A shared at library level with `rw` (positive control) and user B shared at folder level with `rw` on `/Public/` only — no library-level share.

Existing `tests/api/test_file_revert.py` and `tests/api/test_dir_revert.py` cover the library-owner happy path and continue to pass under the new check (because the owner's `rw` on the root resolves to `rw` on every descendant).

Hope this helps. 

Disclaimer: Analysis was guided by claude opus 4.7. Code tested and refined locally by a human. POC available on request.

PR #9068, opened 2026-06-01
fix: allow Unicode characters in share link passwords

### LINKED ISSUE #8623: Share link passwords don't allow non-English letters
Passwords for share links seem to be arbitrarily limited to ASCII, which prohibits using regular, non-English letters.
Examples:

- German umlauts: `äüö`
- Diacritics like in French or Spanish (e.g. `â, é, è, ç`)
- The entire Greek alphabet

When I try using any of those, I get the error message "Password can only contain number, upper letter, lower letter and other symbols.".

1795771535y-cell:
﻿Fixes #8623

## Problem
Share link password validation rejected non-ASCII characters. Passwords with umlauts, diacritics, Greek, CJK etc. were blocked.

Root cause: is_valid_password() used ASCII_RE which rejects any char outside 0x00-0x7F.

## Fix
Replaced ASCII-only check with control-character check. Now allows all printable Unicode characters while still blocking invisible control chars (0x00-0x08, 0x0b-0x0c, 0x0e-0x1f, 0x7f).

seahub/utils/__init__.py

Password length and strength checks remain as additional security layers.

## Testing
Passwords with umlauts, French accents, Greek letters now accepted. Control chars still rejected.

### DIFF
--- a/seahub/utils/__init__.py
+++ b/seahub/utils/__init__.py
@@ -1399,8 +1399,10 @@ ASCII_RE = re.compile(r'[^\x00-\x7f]')
 
 
 def is_valid_password(password):
-
-    return False if ASCII_RE.search(password) else True
+    """Check if password contains only printable characters (any Unicode)."""
+    # Allow any printable or whitespace character, reject control chars only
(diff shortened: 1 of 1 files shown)


1795771535y-cell:
Closing: no maintainer response after extended wait. Thanks for the opportunity.

PR #9073, opened 2026-06-01
Improve context search error message

Napsty:
This PR is related to a question in the Seafile forums (https://forum.seafile.com/t/search-in-wiki-not-supported/25133/2).
When someone wants to use the (context) search in Seafile CE, the current error message just shows "Search not supported". 

image

It's not very clear why, especially for end users (non admins). 

With this PR I hope the understanding of the error message is better.


shuntian:
The issue you reported has been fixed. The search module will no longer be available in future community versions. We've closed your pull request for now. Please feel free to provide feedback if you encounter any problems later.

PR #9343, opened 2026-07-30
fix: initialize group_quota_usage to zero to close #9342

### LINKED ISSUE #9342: Initialization of group_info['group_quota_usage'] to empty string breaks Android Seafile client on non-pro seafile servers
Versions of Android Seafile client release after July 15, 2026 include logic for group quota fields https://github.com/haiwen/seadroid/commits/3a2e6ac9c3f4103eedecf1661b5778ad4c1c347d/app/src/main/java/com/seafile/seadroid2/framework/db/entities/GroupEntity.java and forces the field group_quota_usage to a `long`.

Seahub initializes this field in the backend to an empy string instead. Only if you run a pro version of Seafile, will the field be initialized with an integer.
https://github.com/haiwen/seahub/blob/a2a4c3c46130106374dcec1abcc61e93d14159aa/seahub/api2/endpoints/groups.py#L67-L74

Due to the Android JSON schema parsing, this will cause an exception in the Android app and will fail to retrieve the libraries. 

The fix is simply initializing this field with a zero instead of an empty string. For example I have this in a custom bootstrap script to fix it before booting Seafile.

```bash
sed -i "s/group_info\['group_quota_usage'\] = ''/group_info['group_quota_usage'] = 0/" /opt/seafile/seafile-server-$SEAFILE_VERSION/seahub/seahub/api2/endpoints/groups.py
```

kevinvalk:
See #9342 for details

### DIFF
--- a/seahub/api2/endpoints/groups.py
+++ b/seahub/api2/endpoints/groups.py
@@ -64,7 +64,7 @@ def get_group_info(request, group_id, avatar_size=GROUP_AVATAR_DEFAULT_SIZE):
     if group.parent_group_id != 0:
         group_info['group_quota'] = seafile_api.get_group_quota(group_id)
 
-    group_info['group_quota_usage'] = ''
+    group_info['group_quota_usage'] = 0
     if is_pro_version():
         if is_org_context(request):
             org_id = request.user.org.org_id


freeplant:
Thanks for the PR.

PR #9365, opened 2026-08-05
fix statistics view menu style bug

Michael18811380328:


seafile-dev (approved):
Reviewed the current head. The selector is correctly narrowed to the direct `.more-dropdown-toggle` trigger, so the sizing/flex rules no longer cascade into the dropdown menu item spans. This restores sort-menu labels without changing the intended trigger layout.

PR #9367, opened 2026-08-05
fix column name too long and column icon is distorted

Michael18811380328:


seafile-dev (approved):
Reviewed the hidden-column flex layout on the current head. `flex-shrink: 0` keeps the SVG column-type icon at its intrinsic rendered size while the adjacent `text-truncate` label absorbs overflow. No blocking or warning-level issues found.

PR #9369, opened 2026-08-05
New library settings

llj:


frontend/src/components/dir-view-mode/dir-settings-view/lib-history-setting-panel.js
```diff
@@ -102,49 +102,48 @@ class LibHistorySetting extends React.Component {
...
     this.setState({
-      allHistory: type === 'allHistory' ? true : false,
-      noHistory: type === 'noHistory' ? true : false,
-      autoHistory: type === 'autoHistory' ? true : false,
+      allHistory: type === 'allHistory',
+      noHistory: type === 'noHistory',
+      autoHistory: type === 'autoHistory',
     });
+
+    this.submit();
```
seafile-dev: [Critical] History selection saves old value

Why this matters:
`setLimitDays` schedules the radio/keep-day state updates and immediately calls `submit()`, which reads `this.state`. Selecting “Don't keep history” from “Keep full history”, for example, therefore sends `-1` instead of `0`; the UI then shows the new radio selection even though the persisted setting remains unchanged.

Suggested fix: derive the next days and mode in `setLimitDays` and submit those values directly, or invoke `submit` from the `setState` completion callback after applying one combined state update.

frontend/src/components/dir-view-mode/dir-settings-view/lib-old-files-auto-del-setting-panel.js
```diff
@@ -89,38 +85,36 @@ class LibOldFilesAutoDelSetting extends React.Component {
         isAutoDel: true,
       });
     }
+    this.submit();
```
seafile-dev: [Critical] Auto-deletion switch saves old state

Why this matters:
`updateRadioCheck` calls `submit()` immediately after scheduling `isAutoDel`. Turning automatic deletion on while it is off still submits `0`, and turning it off can submit the old positive day value. The page reports success but the backend retains the opposite setting.

Suggested fix: submit an explicit `nextIsAutoDel`/day value, or pass `submit` as the completion callback of the state update so it reads the committed radio state.

seafile-dev (changes requested):
Verdict: Request changes

Blocking:
- History and auto-deletion radio actions persist the pre-click state because they call `submit()` before React has committed their `setState` updates (see inline comments). This leaves the page showing the selected option while the server retains the previous setting.

Warnings:
- None.

PR #9371, opened 2026-08-06
feat: improve publish UI/UX

Sawtone:


frontend/src/pages/wiki2/side-panel.js
```diff
@@ -45,13 +49,25 @@ class SidePanel extends PureComponent {
...
     const eventBus = EventBus.getInstance();
     this.unsubscribeLibraryToggle = eventBus.subscribe(EXTERNAL_EVENT.ADD_WIKI_LIBRARY_TOGGLE, this.toggleSettingDialog);
+    if (isAdmin && canPublishWiki) {
+      wikiAPI.getPublishWikiLink(wikiId).then((res) => {
+        const { enable_server_render } = res.data;
+        this.setState({
+          canManagePublish: true,
+          enableServerRender: !!enable_server_render,
+        });
+      }).catch(() => {});
```
seafile-dev, line 69: [Warning] Publish entry vanishes on metadata failure

Why this matters:
`isAdmin` and `canPublishWiki` already establish that this user may manage publishing, but `canManagePublish` becomes true only after `getPublishWikiLink` succeeds. A transient network/API failure therefore removes the new Wiki-detail Publish entry completely, so an authorized user cannot publish or unpublish until retrying the whole page.

Suggested fix: set `canManagePublish` from the server-provided permission flags before the request, and use this request only to hydrate `enableServerRender`; on failure retain the entry with a safe default and report the error only if appropriate.

seafile-dev (commented):
Verdict: Comment

Warnings:
- The new Wiki-detail publish entry is gated on a metadata GET that is not required for permission and silently disappears on transient failure. See inline comment.
  Suggested fix: expose the entry from the server-provided permission flags, and treat the GET as best-effort state hydration.
- No automated coverage was added for the new Wiki-detail publish/unpublish state flow; add focused component tests for initial unpublished state, publish success, unpublish success, and metadata-request failure.

PR #9378, opened 2026-08-08
feat: support auto-match icon in Wiki by AI

Sawtone:


frontend/src/components/dialog/add-wiki-dialog/index.js
```diff
@@ -78,24 +97,179 @@ class AddWikiDialog extends React.Component {
...
+    const name = this.state.name.trim();
+    if (!name) return;
+    this.nameBlurTimer = setTimeout(() => {
+      this.nameBlurTimer = null;
+      this.requestSuggestedIcons(name);
+    }, 500);
+  };
+
   handleKeyDown = (e) => {
     if (e.key === 'Enter') {
```
seafile-dev: [Warning] Enter no longer submits the form

Why this matters:
This handler now only prevents the default action; the previous `this.handleSubmit()` call was removed. Keyboard users can no longer confirm either a new Wiki or (in the matching edit handler) an edit with Enter, which regresses existing form behavior.

Suggested fix: call `this.handleSubmit()` after `preventDefault()`. `submitWiki` already blocks submission while AI matching is pending or another submission is in progress.

seafile-dev (commented):
Verdict: Comment

Warnings:
- Pressing Enter no longer submits the add/edit Wiki forms; see the inline comment. The same regression is present in `EditWikiPopover.handleKeyDown`.
  Suggested fix: restore the existing `handleSubmit()` call after `preventDefault()`; its pending-AI and submission guards already prevent premature writes.

seafile-dev (commented):
Verdict: Comment

Warnings:
- Enter 提交行为回归（见行内评论）；编辑 Wiki 的同一处理器也存在相同问题。
  Suggested fix: 在保留现有 pending/submitting 校验的前提下恢复 Enter 到 handleSubmit 的调用。

PR #9379, opened 2026-08-09
Optimize css

HiAriesZhou:


frontend/src/components/common/group-select/index.js
```diff
@@ -66,69 +38,82 @@ class GroupSelect extends Component {
...
+          onKeyDown={Utils.onKeyDown}
+          onClick={this.toggle}
+        >
+          <div className="selected-option">
+            {selectedOptions.length > 0 ?
+              <span className="selected-option-show">
+                {selectedOptions.map(item =>
+                  <span key={item.id} className="selected-option-item">
+                    <span className='selected-option-item-name'>{item.name}</span>
+                    <span className="d-flex align-items-center" onClick={() => {this.props.onDeleteOption(item);}}><Icon symbol="close" /></span>
```
seafile-dev: [Warning] Remove action toggles picker

Why this matters:
Clicking a selected group’s close icon calls `onDeleteOption`, then bubbles to the new parent `onClick={this.toggle}`. The previous handler excluded close-icon events, so removal now also opens or closes the picker in every group-sharing dialog.

Suggested fix: accept the click event in this handler and call `stopPropagation()` before `onDeleteOption(item)`, or restore the previous target exclusion.

frontend/src/pages/dashboard/log-user-selector.js
```diff
@@ -118,78 +98,87 @@ class LogUserSelector extends Component {
...
+          </span>
         </span>
-        {isOpen && (
-          <div className="activity-modifier-selector-container sf-popover-container" ref={ref => this.userSelector = ref}>
-            <ul className="activity-selected-modifiers px-3 py-1 list-unstyled">
-              {selectedItems.map((item, index) => {
+        <Popover
+          isOpen={isOpen}
+          toggle={onToggle}
+          target="log-user-selector-trigger"
```
seafile-dev, line 126: [Warning] Static target breaks parallel selectors

Why this matters:
Every instance uses `log-user-selector-trigger` as its Popover target. Pages such as File Transfer Logs mount several user selectors at once, so the repeated id can resolve a Popover to another trigger and legacy-toggle handlers can affect the wrong selector.

Suggested fix: generate a unique id per instance (or pass the trigger ref/function target) and use it consistently for both the trigger and Popover.

seafile-dev (changes requested):
Request changes: the Popover migration introduces two interaction regressions. See the inline comments for the unique-target and event-propagation fixes.

frontend/src/components/common/group-select/index.js
```diff
@@ -66,69 +38,82 @@ class GroupSelect extends Component {
...
+            fade={false}
+            trigger="legacy"
+          >
+            <SelectOptionGroup
+              selectedOptions={selectedOptions}
+              options={options}
+              onSelectOption={this.props.onSelectOption}
+              searchPlaceholder={searchPlaceholder}
+              noOptionsPlaceholder={noOptionsPlaceholder}
+              onClickOutside={this.toggle}
```
seafile-dev: [Critical] 菜单内部点击会直接关闭选择器

Why this matters:
SelectOptionGroup mounts a document mousedown listener and calls onClickOutside for every event (including clicks on its search input and options). Passing toggle here therefore closes and unmounts the menu before those controls receive their click, so non-modal group selection/search cannot be used.

Suggested fix: restore an outside-click handler that checks both the trigger and Popover/menu DOM before closing, and make closeSelect set isShowSelectOptions to false rather than toggling it.

seafile-dev (commented):
Verdict: Comment

Blocking:
- GroupSelect closes its Popover on every internal document mousedown, so non-modal group option selection and search are unusable (see inline comment). 
  Suggested fix: restore containment-aware outside-click handling and use an explicit close operation rather than a state toggle.

PR #9385, opened 2026-08-11
fix: include OIDC ID token in OAuth logout URL

imwhatiam:


seahub/auth/views.py
```diff
@@ -297,6 +299,21 @@ def logout(request, next_page=None,
     via_oauth = request.COOKIES.get('via_oauth', '')
     oauth_logout_url = getattr(settings, 'OAUTH_LOGOUT_URL', '')
     if (getattr(settings, 'ENABLE_OAUTH', False) or getattr(settings, 'ENABLE_CUSTOM_OAUTH', False)) and via_oauth and oauth_logout_url:
+        if oauth_id_token:
```
seafile-dev, line 302: [Warning] Missing OIDC logout regression coverage

Why this matters:
The existing logout test covers only the normal logged-out page, while this branch changes both session state captured in the OAuth callback and the externally visible redirect URL. A regression in either step can silently omit `id_token_hint` or encode the redirect incorrectly for Dex/Keycloak, which is the behavior this PR is intended to fix.

Suggested fix: add an OAuth-logout test that sets `oauth_id_token` and the `via_oauth` cookie, overrides `OAUTH_LOGOUT_URL`, then parses `Location` and asserts both parameters, URL encoding, and preservation of unrelated configured query parameters. Also retain a no-token case to verify existing fallback behavior.

seafile-dev (commented):
Verdict: Comment

Warnings:
- The OIDC logout redirect path has no automated regression coverage (see inline comment).
  Suggested fix: exercise the callback/session-to-logout redirect contract with an overridden provider URL and parsed query assertions.

PR #9394, opened 2026-08-13
Fix O(n^2) behaviour in recursive directory listing

jgolob:
get_dir_file_recursively() resolved modifier display names inside the per-dirent loop, but all_dirs accumulates across the entire recursive walk -- so every dirent rescanned and rewrote every entry found so far, giving n(n+1)/2 operations. The memo dicts were also re-created on each iteration, so the memcache/db lookups they exist to avoid were re-issued on every pass. The concept of the memcache/db was excellent just needed to be fully implemented.

Here we move the block out of the walk and run it once at the end. Output is unchanged; the same values are computed once instead of n/2 times. The same optimization is applied in the copy in repo_api_tokens/utils.py.

Everywhere else in this codebase the block is already placed after the loop it feeds -- including elsewhere in these same two files. These two recursive copies were the only ones indented one level too deep.

Measured on a ~400,000-dirent library: did not complete in 900s before, 85.7s after. This is the endpoint rclone's seafile backend uses for its recursive listing, so it affects every rclone ls/sync on a large library.

### DIFF
--- a/seahub/api2/views.py
+++ b/seahub/api2/views.py
@@ -2275,7 +2275,7 @@ class UpdateBlksLinkView(APIView):
         url = gen_file_upload_url(token, 'update-blks-api')
         return Response(url)
 
-def get_dir_file_recursively(username, repo_id, path, all_dirs):
+def _collect_dir_file_recursively(username, repo_id, path, all_dirs):
     is_pro = is_pro_version()
     path_id = seafile_api.get_dir_id_by_path(repo_id, path)
     dirs = seafile_api.list_dir_with_perm(repo_id, path,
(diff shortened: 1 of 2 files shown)


jgolob:
If this looks suitable, happy to backport to 13.0 as well

freeplant:
Thanks for the PR. We will give it a review

jgolob:
Just to keep this alive, I did some additional regression testing around output orders and byte-identical. I am happy to share if it helps with the review. :)

seahub/api2/views.py
```diff
@@ -2313,25 +2313,31 @@ def get_dir_file_recursively(username, repo_id, path, all_dirs):
...
+def get_dir_file_recursively(username, repo_id, path, all_dirs):
+    _collect_dir_file_recursively(username, repo_id, path, all_dirs)
+
+    # Resolve modifier display names once, after the whole tree has been
+    # walked.
+    file_list = [item for item in all_dirs if item['type'] == 'file']
+    contact_email_dict = {}
+    nickname_dict = {}
+    modifiers_set = {x['modifier_email'] for x in file_list}
+    for e in modifiers_set:
```
seafile-dev, line 2332: [Warning] 缺少性能回归测试

影响：本次修复依赖“整棵树遍历完成后只解析一次 modifier”。当前没有覆盖多层目录和重复 modifier；以后若解析逻辑被移回递归循环，超大目录库的二次复杂度回归将无法被发现。

修复建议：为两个公开入口添加 mock 测试，断言每个唯一 modifier 仅调用一次解析函数，并校验返回字段与顺序不变。

seafile-dev (commented):
Verdict: Comment

Warnings:
- 递归收集和 modifier 批量解析的拆分符合设计，消除了随 all_dirs 增长而重复扫描的二次复杂度；但尚缺少可防止回归的自动化覆盖。
  Suggested fix: 为两个入口补充多层目录和重复 modifier 的测试，验证每个唯一 modifier 只解析一次且输出保持兼容。

